PR #7, opened 2020-08-13
added higher order energy

fuhaoji:
added higher order energy, changed "particles.py"


ChristopherMayes:
Looks okay, I'll check the units.

PR #9, opened 2021-01-07
Added Lucretia.py

weiyuanlou:
Allows reading and writing Lucretia beam files


ChristopherMayes:
@weiyuanlou The doc strings need to explain inputs and outputs. Like:
https://numpydoc.readthedocs.io/en/latest/format.html

ChristopherMayes:
@weiyuanlou Also use lower case filenames: `lucretia.py`

PR #13, opened 2021-03-31
add correct units to column p

joeduris:
elegant and openpmd expect column p to have units="m$be$nc"


ChristopherMayes:
Thanks @joeduris !

PR #46, opened 2023-08-10
Add methods for slice_statistics and plotting into ParticleGroup

ChristopherMayes:
This adds to `ParticleGroup`:
- `.slice_statistics`
- `.slice_plot`

It also adds the ability to calculate twiss parameters from slices.


ColwynGulliford (approved):
I like the slice key being set automatically

PR #68, opened 2024-07-25
WIP/ENH: wavefront class

ken-lauer:
## Add a Wavefront class

Entirely based on code from @balticfish and many discussions with @ChristopherMayes 

* Automatically handles real space domain padding
* Auto-determines padding values which result in padded array dimensions which:
   1. Are odd integers (per @balticfish, this is required to get balanced/accurate output)
   2. Are efficient for the scipy FFT (https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.next_fast_len.html#scipy.fft.next_fast_len; this offers significant speed-ups)
* Supports
   * Propagation in the Z direction
   * Focusing element in real space
   * Automatic recalculation of real/k-space data (by fft/ifft) on an as-needed basis
* Simple example notebook, modified from @balticfish as well
* Genesis4 notes
    * Propagation is exactly z axis in genesis, propagation can work only with small angle (paraxial approximation for drift propagation)

## Remaining work

- [ ] Refine the API with feedback from @balticfish and @ChristopherMayes 
- [x] Consider avoiding `property` for `.rspace` and `.kspace` as their calculation can take a long time, and IPython auto-completion can automatically evaluate these when you don't want it to (double-check where this was coming from in Jedi; is it still an issue in 2024?)
- [x] Add docstrings to everything
- [ ] Add units to properties that are missing them
- [x] Remove / rename additional properties, depending on utility
- [ ] Remove non-MKS units
    - [ ] `fs` -> `s`
- [x] Standard axis layout of `xyz` is desirable. Is there any benefit to `z,x,y`? Benchmark some operations to verify?
- [ ] Support only meters in the Wavefront class
    - [x] Change initializer to reflect this
    - [x] @balticfish will modify functions with `t`->`m` (`z=c*t`)
- [ ] Define import/export functionality
    - [x] Can we validate with https://github.com/openPMD/openPMD-validator/tree/2.0.X ?
    - [ ] Can we use https://github.com/openPMD/openPMD-api to read/write the data? Possibly an optional dependency?
    - [x] Support reading/writing openpmd-wavefront data
    - [ ] See how this integrates with lume-genesis
- [x] Add some tests
- [x] More after discussions
- [x] Rename `propagate` to `drift`
- [ ] Plotting:
    - [x] XY/ZX/ZY projection: `plot("x", "z")`
    - [x] Plot Re, Im, Abs and Phase
    - [ ] Spectrum `plot_spectrum()` (this is from G.S.; Chris will supply it)
    - [x] Fancy "Wigner" plot (spectrogram-like) (@balticfish has custom code for this)
- [x] Consider moving some methods out to independent functions
   - [x] @ChristopherMayes prefers to have utility functions that aren't baked into the parameter/wavefront classes in this PR
- [x] Add minimal support for polarization - a direction (X stored as `Ex` with openpmd, for example)
    - [ ] Make an issue for future support of generalized polarization, noting that we do not currently have a use for more complicated polarization schemes
- [ ] Ask for feedback from G.S.
- [ ] Rework for clarity - refactored code is now unusable to @balticfish

### DIFF
--- a/pmd_beamphysics/wavefront.py
+++ b/pmd_beamphysics/wavefront.py
@@ -392,7 +392,7 @@ def conversion_coeffs(wavelength: float, dim: int) -> tuple[float, ...]:
     Theta-x, theta-y, omega.
     """
     k0 = 2.0 * np.pi / wavelength
-    return tuple([2.0 * np.pi / k0] * (dim - 1) + [2.0 * np.pi * HBAR_EV_M])
+    return tuple([wavelength] * (dim - 1) + [2.0 * np.pi * HBAR_EV_M])
 
 
 def cartesian_domain(


ChristopherMayes:
- Change `plot` to only make one plot. The user can make multiple calls to `.plot`

ChristopherMayes:
`"xy" implies rspace=True. rspace=False would go with kxky`.

@balticfish could you give a complete list of keys/plots to implement?

ChristopherMayes:
These need some way to set:
- `Wavefront.ranges` 
- `Wavefront._pad.pad` 

These setters are needed for smart propagation (when you resize the grid as the beam grows)

ChristopherMayes:
We also need a method to write the Genesis4 field data from Wavefront (and corresponding test)

pmd_beamphysics/wavefront.py
```diff
@@ -0,0 +1,1972 @@
...
+            float(np.max(z)),  # right
+            float(np.min(kdomain.z) / 2.0),  # bottom
+            float(np.max(kdomain.z) / 2.0),  # top
+        )
+        WT = self.wigner_distribution()
+        plt.imshow(WT, cmap="bwr", extent=extent, aspect="auto")
+        plt.xlabel(f"Time (${kdomain.z_unit_prefix} s$)")
+        plt.ylabel("Photon energy ($eV$)")
+
+    def plot(
```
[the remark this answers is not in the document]
ChristopherMayes: That's fine for the moment, we'll want to iterate. Better will be two separate keys for x and y axes.

pmd_beamphysics/wavefront.py
```diff
@@ -0,0 +1,2084 @@
...
+            if rspace:
+                # See tech note 3.2
+                power_density = 1 / 1e4 * np.abs(mesh_data) ** 2 / (2.0 * Z0)
+                img = plot(power_density, "Power density of mesh (W/cm$^2$)")
+            else:
+                # See tech note 3.6 (coefficient is in 3.4)
+                power_density = (
+                    np.abs(mesh_data) ** 2 / (2.0 * Z0) * self.fft_unit_coeff
+                )
+                img = plot(power_density, "Power density of mesh (J/eV/rad$^2$)")
```
ChristopherMayes, line 1501: Use upright units:
```suggestion
                img = plot(power_density,  "Power density (J/eV/rad$^2$)")
```

pmd_beamphysics/wavefront.py
```diff
@@ -0,0 +1,2084 @@
...
+            img = plot(np.real(mesh_data), title=r"$\Re(\text{mesh})$")
+
+        elif key == "im":
+            img = plot(np.imag(mesh_data), title=r"$\Im(\text{mesh})$")
+
+        elif key == "power_density":
+            if rspace:
+                # See tech note 3.2
+                power_density = 1 / 1e4 * np.abs(mesh_data) ** 2 / (2.0 * Z0)
+                img = plot(power_density, "Power density of mesh (W/cm$^2$)")
```
ChristopherMayes, line 1495:
```suggestion
                img = plot(power_density, "Power density (W/cm$^2$)")

File: pmd_beamphysics/wavefront.py [hunk 4]
```diff
@@ -0,0 +1,2084 @@
...
+            Number of columns for the plot.
+        save : pathlib.Path or str, optional
+            Save the plot to the given filename.
+        xlim : (float, float), optional
+            X axis limits.
+        ylim : (float, float), optional
+            Y axis limits.
+        tight_layout : bool, default=True
+            Set a tight layout.
+        transpose : bool, default=False
```
ChristopherMayes, line 1406: Why is transposing an option here?

pmd_beamphysics/wavefront.py
```diff
@@ -0,0 +1,2084 @@
...
+
+        assert len(sum_axis) == 1
+        sum_axis = sum_axis[0]
+
+        def plot(dat, title: str):
+            # TODO: balticfish will double-check
+            # _z_min, z_max = self.ranges[sum_axis]
+            # dz = self.grid_spacing[sum_axis]
+            # dat = np.sum(dat, axis=sum_axis) * dz / (2.0 * z_max)
+            img = ax.imshow(np.mean(dat, axis=sum_axis), cmap=cmap, extent=extent)
```
ChristopherMayes, line 1465: `sum_axis` implies a sum, but here `np.mean` is called.

For power, it should probably be a sum. Others are questionable.

pmd_beamphysics/wavefront.py
```diff
@@ -0,0 +1,2144 @@
...
+
+
+def conversion_coeffs(wavelength: float, dim: int) -> tuple[float, ...]:
+    """
+    Conversion coefficients to (radians, radians, eV).
+
+    Theta-x, theta-y, omega.
+    """
+    k0 = 2.0 * np.pi / wavelength
+    return tuple([2.0 * np.pi / k0] * (dim - 1) + [2.0 * np.pi * HBAR_EV_M])
```
ChristopherMayes, line 395:
```suggestion
    return tuple([wavelength] * (dim - 1) + [2.0 * np.pi * HBAR_EV_M])
```

PR #74, opened 2024-10-16
Adds 3d GPT field map writing using FieldMesh and CST ASCII file parsing

ColwynGulliford:
This PR pulls in:

 1. Writing 3D field map data to GPT format from FieldMesh.
 2. Reading 3D field map data from CST output ascii files.

1.) This was tested using a 3D corrector field map FieldMesh, loaded directly into Tao, then using the `GPT.from_tao` auto-translation routine to create the corresponding GPT lattice file and associated 3D GPT corrector field map.  Tracking through the map behaves as expected: 1 mrad kick was requested, and this is exactly what GPT gives:

Screenshot 2024-10-16 at 3 28 44 PM

2.) This was tested on two example cases - 
     (a)  975 MHz CERN kicker cavity - for this case the field data was loaded and inspected, the oscillation of the fields was checked using the off-axis expansion of the Maxwell curl equations and the Panofsky-Wenzel theorem.
     (b) 3D DC Gun Field map -  for this case the field data was loaded and inspected, and the divergence of the field checked using the off-axis expansion of the div(E) = 0 equation.


docs/examples/fields/field_conversion.ipynb
ChristopherMayes, line 1: don't include this file

ChristopherMayes (approved):

PR #94, opened 2025-01-14
FIX: show a plot rather than crashing with single particles

ken-lauer:
## Changes

Show a plot rather than crashing when using plot functions on `ParticleGroup` instances holding a single particle.

These are not terribly useful plots, of course, but are better than a crash.

### Screenshots

Before

image

After
image

### DIFF
--- a/pmd_beamphysics/plot.py
+++ b/pmd_beamphysics/plot.py
@@ -284,6 +284,8 @@ def marginal_plot(
     y = particle_group[key2]
 
     if len(x) == 1:
+        bins = 100
+
         if xlim is None:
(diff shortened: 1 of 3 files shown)


ChristopherMayes:
For one particle, we should just plot a point or have better bounds. 20 cm is too big.

ChristopherMayes:
We could have bounds:
- (-1, 1) in base units if the value is 0
- otherwise (0.9, 1.1) times the value or so. bins = 100 or something reasonable to make a spot

ChristopherMayes (approved):

PR #98, opened 2025-02-19
ENH: `pmd_unit` hashability

ken-lauer:
## Changes

I set out to just make `pmd_unit` hashable, but got carried away with some clean-ups along the way.

* `pmd_unit` is now hashable. This means it can be used as a dictionary key or as an element of a set, for example.
* Bonus: added docstrings to some functions missing them
* Bonus: added some type hints and unit tests for `pmd_beamphysics.units`
* Bonus: added support for `*` and `/` operators of 2+ units with a couple simple tests
* Test coverage of `pmd_beamphysics.units` is now pretty good
* Fix mkdocs building after (likely) changes in mkdocstrings-python and our config. This may need to be propagated to other projects.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -81,12 +81,35 @@ plugins:
       default_handler: python
       handlers:
         python:
-          selection:
-            docstring_style: "numpy"
-            inherited_members: false
+          options:
             filters:
               - "!^_" # exclude all members starting with _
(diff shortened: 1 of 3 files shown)


ken-lauer:
Ready for review, but I can't request it (@ChristopherMayes )

ChristopherMayes (approved):

PR #101, opened 2025-05-07
`ParticleGroup` Point Transformation; Rotation

electronsandstuff:
This PR implements the linear point transformation on particle phase space coordinates. That is, given a matrix representing a transformation of the spatial coordinates (r -> M*r), the corresponding canonical transformation is applied to all phase space coordinates. From the requirement that the Lagrangian is unchanged in the translated coordinates and using the definition of conjugate momenta, the momenta coordinates must transform like p -> (M^T)^{-1}*p. A derivation can be found in [Structure and Interpretation of Classical Mechanics](https://tgvaughan.github.io/sicm/chapter005.html).

The following changes are included.
 - New method in `ParticleGroup` to apply point transformation: `linear_point_transform`
 - New method that applies a rotation: `ParticleGroup.rotate`
 - Rotation matrix calculations are moved from `pmd_beamphysics/fields/corrector_modeling.py` to `pmd_beamphysics/utils.py`
 - Unit test for sanity checks on point transformation method
 - New pytest fixture to generate example `ParticleGroup` in `pmd_beamphysics/testing.py`
 - New pytest compatible testing method `assert_pg_close` to check if two `ParticleGroup` objects are approximately equal


ChristopherMayes:
Could we flesh out the routines like:

```python
from typing import Union, Sequence
import numpy as np

def linear_point_transform(self, mat3: Union[np.ndarray, Sequence[Sequence[float]]]) -> None:
    """
    Applies a linear transformation to the particle's spatial coordinates and
    their conjugate momenta.

    The spatial coordinates (x, y, z) are transformed using:
        [x', y', z'] = mat3 @ [x, y, z]

    The conjugate momenta (px, py, pz) are transformed using:
        [px', py', pz'] = inv(mat3.T) @ [px, py, pz]

    Parameters
    ----------
    mat3 : array-like of shape (3, 3)
        A 3x3 transformation matrix. Must be convertible to a NumPy array.
        Coordinates must be ordered as (x, y, z).

    Raises
    ------
    ValueError
        If `mat3` is not a 3x3 matrix.
    LinAlgError
        If the transformation matrix is singular (i.e., not invertible).

    Notes
    -----
    This method modifies the object in-place.
    """
    mat3 = np.asarray(mat3, dtype=float)
    if mat3.shape != (3, 3):
        raise ValueError("mat3 must be a 3x3 matrix.")

    # Stack spatial coordinates and apply the transformation
    pos = np.vstack((self.x, self.y, self.z))
    self.x, self.y, self.z = mat3 @ pos

    # Stack momenta and apply the inverse-transpose transformation
    mom = np.vstack((self.px, self.py, self.pz))
    self.px, self.py, self.pz = np.linalg.solve(mat3.T, mom)
```

The matrix should have some restrictions, right?

ChristopherMayes:
For the rotations, the PALS standard will change the names of these angles to something like `x_rot`, etc.: https://github.com/campa-consortium/pals/blob/5469ee098406fd052874e5478960efe3d8de289e/source/parameters/patch.md?plain=1#L24

Could you use more explicit names like that?

electronsandstuff:
> The matrix should have some restrictions, right?

I think it only has to be non-singular.

electronsandstuff:
I've also been having to make test beams recently and added a little helper method in #109. This PR depends on it now for the benchmark and test fixture if we can merge it first.

ChristopherMayes:
Nice! 

Now for completeness we might as well add an origin default for the individual functions: `def rotate_z(angle, x=0, y=0)`
https://github.com/user-attachments/assets/0a898e88-9765-4da5-a04c-7202102cf9bd

electronsandstuff:
Center defaults changed to zero @ChristopherMayes

PR #109, opened 2025-07-02
`ParticleGroup.from_random_normal()`

electronsandstuff:
Sometimes it is useful for testing/benchmarking to be able to create test distributions without manually building a `ParticleGroup` object yourself. This PR adds the new class method `ParticleGroup.from_random_normal()`. It can be called like this to get an object.

```bash
Python 3.13.3 | packaged by conda-forge | (main, Apr 14 2025, 20:44:30) [Clang 18.1.8 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pmd_beamphysics import ParticleGroup
>>> pg = ParticleGroup.from_random_normal(10000)
>>> len(pg)
10000
```

This returns a normally distributed beam with the requested number of particles. It also allows specifying the mean and covariance and also choosing which time-like coordinate is used.

### DIFF
--- a/pmd_beamphysics/particles.py
+++ b/pmd_beamphysics/particles.py
@@ -227,8 +227,8 @@ class ParticleGroup:
     def from_random_normal(
         cls,
         n_particle: int,
-        mean: np.ndarray | None = None,
-        cov: np.ndarray | None = None,
+        mean: Union[np.ndarray, None] = None,
+        cov: Union[np.ndarray, None] = None,
         species: str = "electron",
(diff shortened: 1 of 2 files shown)


ChristopherMayes:
I'd like to avoid constructors for ParticleGroup that have hard-coded default params as part of ParticleGroup. We can have the general method, but the user must always be asked to supply the numbers. 

You could make the covariance matrix, etc. in the tests, or have a helper function in tests to make the particles.

electronsandstuff:
Closing and moving to external function in #101 while we think of interface.

PR #111, opened 2025-07-17
Add longitudinal Wakefields

ChristopherMayes:


pmd_beamphysics/wakefields/resistive_wall.py
```diff
@@ -0,0 +1,1574 @@
...
+
+        The dimensionless scaled wake is Ŵ(z/s₀) = W(z) / W₀.
+        """
+        W0_round = c_light * Z0 / (np.pi * self.radius**2)
+        if self.geometry == "round":
+            return W0_round
+        else:  # flat
+            return W0_round * np.pi**2 / 16
+
+    def _extract_z_weight(self, particle_group_or_z, weight=None):
```
electronsandstuff: This is more of a `ParticleGroup` helper function. Maybe it should go in the base file or in some sort of utilities file?

pmd_beamphysics/wakefields/base.py
```diff
@@ -0,0 +1,1117 @@
...
+        Returns
+        -------
+        W : float or np.ndarray
+            Wakefield value [V/C/m]. Returns 0 for z > 0.
+        """
+        if self._wakefield_func is not None:
+            return self._wakefield_func(z)
+
+        # Numerical cosine transform
+        from scipy.integrate import quad
```
electronsandstuff: Move imports to top of file

pmd_beamphysics/particles.py
```diff
@@ -1184,6 +1185,76 @@ def slice_plot(
         if return_figure:
             return fig
 
+    def wakefield_plot(
```
electronsandstuff: Not the worst thing, but I will point out that this is the only method coupling `ParticleGroup` with the wakefields module. Removing this just means the user calls
```
wakefield_plot(pg, wf)
```
instead of 
```
pg.wakefield_plot(wf)
```
but reduces coupling between the sections of the code.

pmd_beamphysics/wakefields/base.py
```diff
@@ -0,0 +1,1117 @@
...
+
+    def _self_kick_value(self) -> float:
+        """
+        Return the self-kick value W(0⁻).
+
+        Subclasses should override this for efficiency.
+        """
+        return self.wake(-1e-15)
+
+    def apply_to_particles(
```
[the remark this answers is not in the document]
ChristopherMayes: This is mostly for analysis convenience. I don't plan on implementing solvers in this package. Note that simply multiplying by a length is what is often done in practice. For example. at LCLS the resistive wall wake is applied in one place with a 1 km length! I can add a note that this is simple.

pmd_beamphysics/wakefields/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+    "WakefieldBase",
+    "Pseudomode",
+    "PseudomodeWakefield",
+    "TabularWakefield",
+    "ImpedanceWakefield",
+    # Resistive wall
+    "ResistiveWallWakefieldBase",
+    "ResistiveWallWakefield",
+    "ResistiveWallPseudomode",
+    "pseudomode",  # Legacy alias
```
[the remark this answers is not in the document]
ChristopherMayes: Thanks, removed.

pmd_beamphysics/wakefields/resistive_wall_impedance.py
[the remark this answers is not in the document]
ChristopherMayes: I did a major refactor, removing a lot of redundant code.

pmd_beamphysics/wakefields/resistive_wall.py
ChristopherMayes: Removed, now the user needs to always supply `z` and `weight`.

pmd_beamphysics/wakefields/resistive_wall/base.py
```diff
@@ -0,0 +1,721 @@
...
+from enum import Enum
+
+import numpy as np
+from scipy.integrate import quad, quad_vec
+
+from ...units import c_light, Z0
+from ..base import WakefieldBase
+
+
+class Geometry(str, Enum):
```
electronsandstuff: My thought was more like this:

```
class Geometry:
    ...

class GeometryRound(Geometry):
    radius: float

    def W0(self):
        return c_light * Z0 / (np.pi * self.radius**2)

class GeometryFlat(Geometry):
    gap: float

    def W0(self):
        return c_light * Z0 /  self.gap**2 * np.pi / 4
```
That way the parameter names are specialized to the type of geometry and potentially more complicated things that need stuff more than a radius could be accomodated later on.

Then
```
class ResistiveWallWakefieldBase(WakefieldBase):
    conductivity: float
    relaxation_time: float
    geometry: Geometry = GeometryRound()

    def W0(self):
        return self.geometry.W0()
```

electronsandstuff (approved):

PR #133, opened 2026-02-20
ENH/REF: split up plotting + add Bokeh plotting

ken-lauer:
## Overview

* Separates out plot data preparation and plotting.
* Adds Bokeh plot equivalents of the existing Matplotlib ones.
   * Bokeh will remain an optional dependency for the library.
* Plot functions now accept `backend="mpl"` or `backend="bokeh"`, similar to how we do it in PyTao.
   * The default remains `"mpl"`, for back-compat and such. Users can configure this as the `BEAMPHYSICS_PLOT` environment variable (`"bokeh"` or `"mpl"` for now

### Attribution

I did the initial prepare/plot split and the Bokeh marginal plot, then added Claude into the mix to tackle the remainder following the pattern. Many fixes on top of that, and also bringing in some PyTao notebook handling tools I wrote a while back.


ken-lauer:
This is getting closer and probably should have some eyes on it. Take a look at `docs/examples/plot_backend_comparison.ipynb` for a side-by-side.
There is also a small addition to the `docs/examples/plot_examples.ipynb` notebook.
(nudge @ChristopherMayes )

ChristopherMayes:
Nice! some math labels don't render properly:
image


Also, perhaps use `beamphsysics.set_default_backend` in the notebook (correct spelling in the notebook.

ken-lauer:
Using `set_default_backend` in the example now (with the right spelling, hopefully)
For the legend stuff, it seems like Bokeh doesn't actually support LaTeX there: reference https://github.com/bokeh/bokeh/issues/12844

There are some workarounds, but they're not so pretty (like putting a `div` off to the side and making your own legend)
I think we might need to just show basic, non-LaTeX text there for starters...

ken-lauer:
This will need a very careful rebase/merge after #142 made it in 😢

PR #140, opened 2026-04-21
PERF: lazy imports at the top-level

ken-lauer:
## Background

`beamphysics` currently takes almost 1 second to import on my machine.
The majority of that time is spent importing `scipy` and `matplotlib`, which are from nested imports in the library.

Until the user actually needs other bits of beamphysics, lazy importing is the way to go.

### Changes

That is what this PR implements:

1. For type checking, there's a block at the top that allows pyright/pylance (VS Code) to see what's in the module
2. For access of those items, `__getattr__` handles the imports dynamically. The first import incurs a cost, but subsequent imports are cached automatically (via `globals()` patching)
3. `__dir__` is a bit of an extra bit not truly required, but it's supported by the module interface, and should show everything that is/would be imported

While a bit more complicated than the main branch, this shouldn't be so complicated to update with future top-level imports.

## Performance

import_times-log

Yes, that's less than a millisecond for just the import.
Our commonly-used `beamphysics.units` takes much less time, too:

import_times-units

(Sorry, I switched scales there. I think the point still stands)

## What's next?

I'm onto distgen next, as it takes over a second to import thanks to `pint`!

### DIFF
--- a/beamphysics/__init__.py
+++ b/beamphysics/__init__.py
@@ -1,14 +1,31 @@
-from .fields import FieldMesh
-from .particles import ParticleGroup, single_particle
-from .readers import particle_paths
-from .status import ParticleStatus
-from .wavefront import Wavefront, WavefrontK
-from .writers import pmd_init
+from __future__ import annotations
+
+import typing as _typing
+
+if _typing.TYPE_CHECKING:
+    from .fields import FieldMesh
+    from .particles import ParticleGroup, single_particle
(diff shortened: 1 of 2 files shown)


ken-lauer:
Edit: here was an unrelated issue that made it look like this change wouldn't be compatible with pydantic. It is compatible and I don't think there will be an issue going forward with this.

ChristopherMayes (approved):

PR #142, opened 2026-06-06
Overhaul `pmd_unit` — expand named units, harden parser, value-based equality

ChristopherMayes:
## Summary

Changes to `beamphysics/units.py`:

1. **New named units**: `Hz` and `deg` (alias of `degree`). `T/m` is *not* added as a named unit — see §1.
2. **Bug fix**: `electric_potential` dimension had length and mass exponents swapped.
3. **Improved `pmd_unit.simplify`**: float-tolerant matching, deterministic tie-breaking, and a new compound-unit search that handles `a*b` / `a/b` combinations.
4. **Round-trip safety in `simplify`**: the compound search no longer emits symbols that re-parse to a different unit, and an SI prefix is no longer glued onto a dimensionless symbol. See §4.
5. **Global-state fix**: parsing a compound unit string no longer mutates the shared `NAMED_UNITS` entries in place. See §5.
6. **SI-prefix parsing**: `pmd_unit` now parses prefixed symbols (`keV`, `mm`, `mrad`, `GHz`, …), making `simplify`'s prefixed output round-trip. See §6.
7. **`sqrt_unit` dimension fix**: `sqrt(m)` was collapsing to dimensionless instead of `m^0.5`, mis-typing `x_bar`/`px_bar`/`y_bar`/`py_bar`. See §7.
8. **`cov_` key fix**: `pg_units("cov_charge__x")` raised `KeyError`. See §8.
9. **New named unit `Ω` (ohm)**, with an `Ohm` ASCII alias — impedance, used by the wakefields module. See §9.
10. **Symbol canonicalization & input hardening**: whitespace around operators and inside `sqrt(...)` is normalized; malformed compound inputs (`m*`, `/m`, `m//s`, `m**s`, `\sqrt{m}`) and negative `unitSI` are rejected. See §10.
11. **`power_unit` distributes across compounds**: `(eV*s)^2 → eV^2*s^2` (not the malformed `eV*s^2`). See §11.
12. **`pmd_unit.to_tex()` for plot labels**: structural translation from stored symbol to matplotlib-mathtext; `mathlabel` consumes it at the boundary. See §12.
13. **Value-based equality (openPMD spec)**: `__eq__` compares `(unitDimension, unitSI)`, not the presentational symbol. `pmd_unit("1/s") == pmd_unit("Hz")`. See §13.
14. **API rounding-out**: `__pow__` (`pmd_unit("m") ** 2`), `compatible_with()` (pure-dimension check), and `conversion_factor_to()` (SI factor with dimension guard). `read_dataset_and_unit_h5` now uses these instead of an ad-hoc `assert`, and `norm_emit_4d` is stored as `m^2` rather than `m*m`. See §14.
15. **Removed the redundant `vel/c` named unit**: it was a velocity-dimensioned clone of `c` (`unitSI = c_light`), not a dimensionless β; the legacy `c_light` alias now points at `c`. See §15.
16. **`divide_units` emits round-trip-safe symbols for compound divisors**: dividing by `eV/c` now yields `.../(eV/c)` instead of `.../eV/c` (which the left-associative parser regroups). Same root cause as §4a, fixed at the arithmetic layer. See §16.
17. **Unicode look-alike characters accepted on parse**: both versions of `Ω` (OHM SIGN U+2126 and GREEK CAPITAL OMEGA U+03A9) and of `µ` (GREEK SMALL MU U+03BC and MICRO SIGN U+00B5) parse as the characters the unit tables use. See §17.
18. **Parenthesized grouping in the symbol grammar**: `V/(eV/c)` and `(eV*s)^2` now parse, removing the structural root cause behind the §4a/§11/§16 workarounds. See §18.
19. **Package-wide unit-usage sweep**: a review of every `units` consumer in the package found and fixed several bugs (a float-invariant hole in named units, a `sqrt(m)*sqrt(m)` parse bug, plot labels that mis-prefixed compound symbols, an HDF5 bytes/str symbol mismatch) and applied the new API where ad-hoc code duplicated it. Non-unit bugs found by the same review are split into a follow-up PR. See §19.
20. **Single recursive parser**: `_parse_single_unit` is merged into `from_symbol`, which now recurses on tokens, group/sqrt inners, and power bases. This removes the duplicated lookup/sqrt logic that had already drifted (`sqrt(kg*m)` parsed alone but failed inside `sqrt(kg*m)*s`). See §20.

An independent `pint`-based test suite (`tests/test_units_pint.py`) cross-checks the unit definitions, parser, and `pg_units` table against the `pint` library. `pint` is a **test-only** dependency; the package does not import it.

All tests pass (1796 passed, 1 xpassed), including new regression tests for every fix above.

---

## 1. New named units

Added to `NAMED_UNITS`:

| Symbol | `unitSI` | `unitDimension`            | Notes |
|--------|----------|-----------------------------|-------|
| `Hz`   | 1        | `(0, 0, -1, 0, 0, 0, 0)`    | Inverse seconds. Cannot be built from `*`/`/` of existing names, so naming is required for `pmd_unit("Hz")` to resolve. |

Added to `known_unit` as an alias:

| Symbol | Aliases  | Notes |
|--------|----------|-------|
| `deg`  | `degree` | Common short form seen in element headers (`rf_phase_deg`, `theta0_deg`). |

These cover the units `Hz` (`Bfreq`, `rf_frequency`) and `deg` (`rf_phase_deg`, `theta0_deg`) that appear in common element headers.

`T/m` (`b1_gradient`) is *not* added as a named unit — the improved `simplify()` (see §3) reconstructs `"T/m"` from `T` and `m` automatically, and the parser already handles `pmd_unit("T/m")` through its compound-symbol parsing.

## 2. Bug fix: `electric_potential` dimension

A volt is `J/C = kg·m²·s⁻³·A⁻¹`. The previous entry had length and mass swapped:

```diff
- "electric_potential": (1, 2, -3, -1, 0, 0, 0),  # wrong: L and M swapped
+ "electric_potential": (2, 1, -3, -1, 0, 0, 0),
```

### Verification

Cross-checked against neighbouring dimensions, which are correct:
- `electric_field` (`V/m`): `(1, 1, -3, -1, 0, 0, 0)` — one less length than V.
- `W = V·A = J/s`:        `(2, 1, -3,  0, 0, 0, 0)` — one less current than V.

So `V` must have length 2 and mass 1.

### Bonus

With the dimension corrected, `simplify()` now resolves `V*A → W` and `J/C → V` correctly via the new compound search.

### Compatibility note

Any HDF5 file written by an earlier version that used the old `unitDimension = (1, 2, -3, -1, 0, 0, 0)` for volts is now inconsistent with this package's view of `electric_potential`. `read_unit_h5` still reads them (it trusts the file attributes verbatim), but dimension- compatibility checks against `dimension("electric_potential")` would flag them as mismatched.

### Problems with the previous implementation

- Used `factor == 1` and `factor in SHORT_PREFIX` (dict lookup). Float values produced by unit arithmetic (e.g. `(eV/c)*c`) rarely hit those exact comparisons, so simplification silently failed.
- Only matched a *single* named unit's dimension. Even when both factors of a compound were named (`T*m`, `J*s`, `V*A`), nothing simplified.
- First-match ordering meant the output depended on `NAMED_UNITS` ordering when several entries share a dimension (`eV` vs `J`, `eV/m` vs `J/m`, ...).

### New behavior

Two-pass search:

1. **Atomic match.** Find named units whose dimension matches and whose `unitSI` matches the target either exactly or up to a single SI prefix factor (`SHORT_PREFIX_FACTOR`). Comparisons use `math.isclose(..., rel_tol=1e-9)`.
2. **Compound match.** If no atomic match exists, search every named pair `(a, b)` (both with nonzero dimension) for an exact `a*b` or `a/b` whose dimension and `unitSI` match the target.

Among candidates the simplest is returned, using the sort key `(has_operators, length, lexicographic)` — so atomic beats compound, shorter beats longer, and ties are deterministic.

If nothing matches, `self` is returned (unchanged behavior).

### New helpers (module-private)

- `_SIMPLIFY_REL_TOL = 1e-9`
- `_symbol_complexity(symbol)` — sort key for candidate ranking.
- `_closest_prefix(factor)` — float-tolerant SI prefix lookup.
- `_best_atomic_match(target_dim, target_si, named_units)`
- `_best_compound_match(target_dim, target_si, named_units)`

### Examples

```python
>>> ((pmd_unit("eV/c") * pmd_unit("c")).simplify())   # was failed silently
pmd_unit('eV', 1.602176634e-19, (2, 1, -2, 0, 0, 0, 0))

>>> (pmd_unit("T") * pmd_unit("m")).simplify()        # NEW: compound search
pmd_unit('T*m', 1, (1, 1, -2, -1, 0, 0, 0))

>>> (pmd_unit("V") * pmd_unit("A")).simplify()        # NEW: needs V fix + compound
pmd_unit('W', 1, (2, 1, -3, 0, 0, 0, 0))

>>> ((pmd_unit("T") / pmd_unit("m") / pmd_unit("m")) * pmd_unit("m")).simplify()
pmd_unit('T/m', 1, (-1, 1, -2, -1, 0, 0, 0))
```

### Performance

The compound search builds a `dim -> [pmd_unit]` index once and iterates through `~N` named units, doing dict lookups for compatible partners. Worst-case cost is `O(N²)` for `N ≈ 30` named units, which is trivial.

## 4. Round-trip safety in `simplify`

The symbol parser is **left-associative**: `"a/b/c"` means `(a/b)/c`. (At this point in the PR it had no parenthesized grouping; §18 adds that later.) The first draft of the compound search and the prefix logic could both produce a symbol *string* whose stored `(unitSI, unitDimension)` were correct but which re-parsed to a *different* unit. Two cases were fixed.

### 4a. Division by a compound divisor

When the divisor `b` already contains an operator, `f"{a}/{b}"` is wrong:

```python
# before: dim/si forced to target, but the SYMBOL lies
>>> (pmd_unit("V/m") / pmd_unit("W")).simplify()
pmd_unit('T/J/m', 1.0, (-1, 0, 0, -1, 0, 0, 0))
>>> pmd_unit('T/J/m')                       # 'T/J/m' == (T/J)/m, not T/(J/m)
pmd_unit('T/J/m', 1.0, (-3, 0, 0, -1, 0, 0, 0))   # different dimension!
```

Fix: `_best_compound_match` now skips any divisor whose symbol contains `*` or `/`. The `a*b` form is unaffected — a product chain is associative under the parser's grouping, so `a*b` always round-trips.

```python
>>> (pmd_unit("V/m") / pmd_unit("W")).simplify()
pmd_unit('V/m/W', 1.0, (-1, 0, 0, -1, 0, 0, 0))   # round-trips
```

### 4b. SI prefix on a dimensionless symbol

A prefix on a dimensionless unit produced a bare prefix glued to the empty symbol, which is wrong or unparseable:

```python
# before
>>> pmd_unit("ratio", 1e-3, (0,)*7).simplify()
pmd_unit('m', 0.001, (0, 0, 0, 0, 0, 0, 0))   # 'm' re-parses as METERS
>>> pmd_unit("ratio", 1e3, (0,)*7).simplify()
pmd_unit('k', 1000.0, (0, 0, 0, 0, 0, 0, 0))  # 'k' is unparseable
```

Fix: `_best_atomic_match` no longer applies a prefix to a dimensionless unit. Only an exact (unprefixed) dimensionless match is accepted; otherwise the search falls through to the compound pass or returns `self`.

A brute-force check over every `a*b` / `a/b` of the named units confirms that `simplify()` never turns a round-tripping unit into a non-round-tripping one.

## 5. Global-state fix: parsing no longer mutates `NAMED_UNITS`

**Pre-existing bug, independent of this PR** (also present on `master`), but surfaced by the new `simplify` tests, which are the first to rely on `NAMED_UNITS` integrity within a process.

`pmd_unit._parse_single_unit` returned the **shared** cached unit object, and `from_symbol` then reassigned `result._unitSymbol`. Because `rad` is an identity unit (`unitSI == 1`, dimensionless), `multiply_units` returns its operand unchanged, so parsing a string like `"m*rad"` rewrote the **global** `"m"` entry's symbol to `"m*rad"` — silently corrupting every later `simplify()` call in the same process:

```python
>>> import beamphysics.units as U
>>> pmd_unit("m*rad")
>>> [u.unitSymbol for u in U.NAMED_UNITS if u.unitSI == 1][:4]
['', 'degree', 'rad', 'm*rad']   # 'm' was clobbered
```

Fix: `_parse_single_unit` returns a copy of the cached unit instead of the shared object (mirroring what `from_symbol` already does for direct lookups). Parsing is now side-effect-free.

## 6. SI-prefix parsing

Previously `pmd_unit` could not parse prefixed symbols — `pmd_unit("keV")`, `"mm"`, `"mrad"`, `"GHz"` all raised `Unknown unitSymbol`. This was also an asymmetry bug: `simplify()` *emits* prefixed atomic symbols (`"mg"`, `"kV"`), so e.g. `(g*g/kg).simplify()` produced `"mg"`, which then failed to re-parse.

Fix: `_parse_single_unit` gained an SI-prefix fallback as its last step. After exact lookup / `sqrt` / `^power` fail, it tries each prefix (longest first, so `da` beats `d`) and accepts `prefix + remainder` **only when the remainder is itself a known unit**. This is the exact inverse of what `simplify()` emits, so prefixed output now always round-trips.

```python
>>> pmd_unit("MeV").unitSI, pmd_unit("MeV").unitDimension
(1.602176634e-13, (2, 1, -2, 0, 0, 0, 0))
>>> (pmd_unit("g") * pmd_unit("g") / pmd_unit("kg")).simplify()
pmd_unit('mg', 1e-06, (0, 1, 0, 0, 0, 0, 0))   # now re-parses correctly
```

Rules (by design):

- A bare prefix is **not** a unit: `"k"`, `"M"`, `"da"`, `"k1"` still raise.
- Lookup happens before the fallback, so prefix letters that are also unit symbols keep their unit meaning: `c` = speed of light (never centi), `T` = tesla, `g` = gram, `cd` = candela.
- `Tm` parses as tera-meter. Tesla·meter must be written `T*m` (the package already requires `*` for products), so adjacency like `Tm` is a user error and is intentionally not second-guessed.

## 7. `sqrt_unit` dimension fix

`sqrt_unit` computed the result dimension with integer floor division (`x // 2`), so `sqrt(m)` (dimension `(1, 0, …)`) collapsed to `(0, 0, …)` — **dimensionless** instead of `m^0.5`. This mis-typed the `pg_units` entries built from it: `x_bar`, `px_bar`, `y_bar`, `py_bar`, which `particles.py` documents as having units of `sqrt(m)`.

```diff
- dim = tuple(x // 2 for x in u.unitDimension)   # 1 // 2 == 0
```

Fix: `sqrt_unit` now delegates to `power_unit(u, 0.5)`, which scales the exponents by `0.5` (so `m` → `m^0.5`) while preserving the `\sqrt{…}` display symbol. To make fractional exponents survive the round trip through the constructor, `Dimension` is now uniformly a 7-tuple of floats and `make_dimension` coerces to `float` (instead of `int`). This also matches the openPMD standard, which specifies `unitDimension` as "array of 7 *(float64 / REAL8)*". `pg_units("x_bar")` is now `(0.5, 0.0, …)`, and `x_bar**2` correctly gives length.

## 8. `cov_` key fix in `pg_units`

`pg_units` parsed covariance keys with `key.strip("cov_")`. `str.strip` removes any leading/trailing characters in the set `{c, o, v, _}` — not the prefix `"cov_"` — so a subkey beginning with one of those letters was mangled:

```python
>>> pg_units("cov_charge__x")   # before
KeyError: 'harge'               # "charge" -> "harge"
```

Fix: use `key.removeprefix("cov_")`, which strips exactly the prefix.

## 9. New named unit: `Ω` (ohm)

The wakefields module documents impedance in `[Ohm/m]` and plots `Z(k)` in `Ω/m`, but `pmd_unit("Ω")` / `pmd_unit("Ohm")` raised `Unknown unitSymbol` — so those unit strings were never expressible through the units machinery.

Added `Ω` to `NAMED_UNITS` with dimension `(2, 1, -3, -2, 0, 0, 0)` (ohm = V/A), plus an `Ohm` ASCII alias in `known_unit` (mirroring the `deg`/`degree` precedent). Consequences:

```python
>>> pmd_unit("Ω/m").unitDimension       # impedance per length
(1.0, 1.0, -3.0, -2.0, 0.0, 0.0, 0.0)
>>> pmd_unit("kΩ").unitSI                # prefixes work
1000.0
>>> (pmd_unit("V") / pmd_unit("A")).simplify()
pmd_unit('Ω', 1, (2.0, 1.0, -3.0, -2.0, 0.0, 0.0, 0.0))
```

Only `Ω` was added: it is the one missing unit with actual in-repo usage. Other absent SI units (`N`, `Pa`, `Wb`, `F`, `S`, ...) are not referenced anywhere and force is already covered by the named `eV/m` / `J/m`.

A regression test also asserts that every unit string the wakefields module documents/plots (`V/C/m`, `Ohm/m`, `Ω/m`, `V/pC/m`, `MV/m`, `1/mm`, ...) is a valid `pmd_unit` symbol, keeping those annotations consistent with the units system.

## 10. Symbol canonicalization & input hardening

The parser and constructor now reject several malformed inputs that previously silently parsed via the identity unit, and canonicalize whitespace on the way in so trivially-different spellings of the same unit hash and compare equal.

### Whitespace normalization

Whitespace around operators (`*`, `/`), around `^` exponents, and inside `sqrt(...)` is stripped during parsing, so the **stored** `unitSymbol` is whitespace-free:

```python
>>> pmd_unit("eV / c").unitSymbol
'eV/c'
>>> pmd_unit("kg * m / s").unitSymbol
'kg*m/s'
>>> pmd_unit("m ^ 2").unitDimension
(2.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0)
>>> pmd_unit("sqrt( m )").unitSymbol
'sqrt(m)'
```

Combined with the value-based equality in §13, this means `pmd_unit("eV / c") == pmd_unit("eV/c")` and they share a hash.

### Malformed inputs rejected

Previously `"m*"` silently parsed as `m * 1 = m` because the empty operand resolved to the identity unit; `"/m"` became `1/m`; `r"\sqrt{m}"` was accepted by an experimental LaTeX branch. All now raise `ValueError`:

```python
>>> pmd_unit("m*")        # empty operand around '*'
ValueError: Malformed unit symbol 'm*': empty operand around an operator.
>>> pmd_unit("m**s")      # double operator -> empty operand between
ValueError: ...
>>> pmd_unit(r"\sqrt{m}") # only the sqrt(...) form is accepted
ValueError: Unknown unitSymbol: ...
```

The `\sqrt{...}` parser branch was removed: only the canonical `sqrt(<inner>)` form is accepted on input, and `sqrt_unit` / the sqrt parser canonicalize whitespace inside to that same form on output.

### Negative `unitSI` rejected

The openPMD standard defines `unitSI` as a conversion factor to SI. A negative scale would invert the sign of every value an instrument writes through the unit, so the constructor now rejects it (zero is still accepted as the sentinel that triggers symbol lookup):

```python
>>> pmd_unit("bogus", unitSI=-1.0, unitDimension="1")
ValueError: unitSI must be non-negative (got -1.0). The openPMD standard
defines unitSI as a conversion factor to SI.
```

## 11. `power_unit` distributes across compound symbols

Writing `f"{symbol}^{n}"` for a compound symbol produces the wrong unit: `"eV*s^2"` parses as `eV*(s^2)`, not `(eV*s)^2`. `power_unit` distributes the exponent across each token and re-emits, so the symbol always round-trips and matches the dimension (the distributed `eV^2*s^2` is preferred over the also-valid `(eV*s)^2` as the conventional spelling):

```python
>>> power_unit(pmd_unit("eV*s"), 2)
pmd_unit('eV^2*s^2', 2.567e-38, (4.0, 2.0, -2.0, 0.0, 0.0, 0.0, 0.0))
>>> power_unit(pmd_unit("m^2*s^-1"), 3)
pmd_unit('m^6*s^-3', 1.0, (6.0, 0.0, -3.0, 0.0, 0.0, 0.0, 0.0))
>>> power_unit(pmd_unit("m^2*s^2"), 0.5)   # halving the compound is also correct
pmd_unit('m*s', 1.0, (1.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0))
```

`sqrt_unit` now uses the canonical `sqrt(<inner>)` form (no LaTeX, no whitespace) so it round-trips through the parser. Internal helpers added: `_tokenize_compound`, `_join_compound`, `_distribute_power_in_symbol`, `_combine_token_power`, `_format_power`, `_is_atomic_symbol`.

`_best_compound_match` was tightened in step: both the numerator and divisor of an emitted `a*b` / `a/b` must be **atomic** (no top-level `*`/`/`). A compound numerator parses correctly but reads ambiguously to a human (`"J/m*s"` left-associates to `(J/m)*s`, but looks like `J/(m*s)`); restricting to atomic operands keeps the emitted symbol unambiguous.

## 12. `pmd_unit.to_tex()` for plot labels

`pmd_unit` now has a `to_tex()` method that renders the stored symbol as a matplotlib-mathtext fragment. The translation is **purely structural** — no algebraic simplification, no reordering: each atomic name is wrapped in `\mathrm{...}`, `*` becomes `{\cdot}`, `/` stays, `^N` becomes `^{N}`, `^(n/d)` becomes `^{n/d}`, and `sqrt(X)` becomes `\sqrt{...}` (recursing on the inner expression).

```python
>>> pmd_unit("eV/c").to_tex()
'\\mathrm{eV}/\\mathrm{c}'
>>> pmd_unit("kg*m/s").to_tex()
'\\mathrm{kg}{\\cdot}\\mathrm{m}/\\mathrm{s}'
>>> pmd_unit("sqrt(kg*m)").to_tex()
'\\sqrt{\\mathrm{kg}{\\cdot}\\mathrm{m}}'
>>> pmd_unit("m*s/m").to_tex()             # preserves stored structure
'\\mathrm{m}{\\cdot}\\mathrm{s}/\\mathrm{m}'
```

`beamphysics/labels.py::mathlabel` was updated to consume `to_tex()` when wrapping a `pmd_unit` (or a str that parses to one). This restores matplotlib's `\sqrt{...}` glyph rendering for `sqrt_unit(pmd_unit("m"))` quantities like `x_bar`, which would otherwise have shown the literal `sqrt(m)` text after the §10 sqrt-canonicalization removed the LaTeX-wrapped storage form.

Two edge cases in `mathlabel` are handled explicitly (both caught in review of an earlier draft of this PR):

- **Unparseable unit strings** (e.g. `units="foo*bar"`) fall back to `\mathrm{...}` with `*` still replaced by `{\cdot}`, matching the pre-`to_tex` behavior.
- **Dimensionless units in the non-TeX branch**: a `pmd_unit` object is always truthy, so the `tex=False` branch gated on the object would emit `"x ()"` for a unit with an empty symbol. It now gates on `str(units)`, so a dimensionless unit adds no annotation (matching the TeX branch and the pre-PR behavior).

## 13. Value-based equality (openPMD spec)

The openPMD standard defines a unit by its `unitSI` and `unitDimension`; the symbol is presentational. `__eq__` now follows that convention — two units compare equal when they share the same dimension and (within a small relative tolerance) the same SI factor, regardless of how their symbols were spelled:

```python
>>> pmd_unit("1/s") == pmd_unit("Hz")              # different syntax, same unit
True
>>> pmd_unit("s^-1") == pmd_unit("Hz")
True
>>> pmd_unit("eV/c") * pmd_unit("c") == pmd_unit("eV")   # arithmetic round-trip
True
>>> pmd_unit("J") == pmd_unit("eV")                # same dim, different SI scale
False
```

Both `__eq__` and `__hash__` derive `unitSI` from a single canonical key, `_canonical_unitSI`, which quantizes the factor to 12 significant figures. This absorbs the ULP-level drift of arithmetic chains like `(eV/c) * c` (so they still satisfy `== eV`) while giving a **transitive** equality — unlike a bare `math.isclose`, which is not transitive and cannot be made consistent with any finer-than-dimension hash (two values within tolerance can straddle a rounding boundary, breaking `a == b ⟹ hash(a) == hash(b)`).

`__hash__` therefore keys on `(unitDimension, _canonical_unitSI(unitSI))`. Hashing on the dimension *alone* would be correct but would funnel every same-dimension unit (`eV`, `J`, `keV`, `MeV`, …) into a single bucket; including the canonical SI key keeps distinct scales in distinct buckets while preserving the invariant. Sets and dicts de-duplicate correctly:

```python
>>> len({pmd_unit("1/s"), pmd_unit("Hz")})            # equal -> de-duplicated
1
>>> len({hash(pmd_unit(u)) for u in ("eV", "J", "keV", "MeV")})  # distinct buckets
4
```

### Downstream impact

This is the one change in this PR that can alter downstream behavior without raising an error: any code using `pmd_unit` objects as dict keys or in sets where symbol-distinct-but-value-equal units (`"1/s"` vs `"Hz"`, `"m*m"` vs `"m^2"`) previously stayed separate will now silently merge them. This is the openPMD-correct semantics, but it deserves a prominent line in the release notes/changelog.

### Why not auto-simplify on construction?

An earlier draft considered canonicalizing the stored symbol on construction (so `pmd_unit("1/s")` would store as `Hz`). Rejected: it loses user-supplied structure that `to_tex()` and downstream consumers may want to preserve (`pmd_unit("m*s/m").unitSymbol == "m*s/m"`), pays the `simplify()` cost on every construction, and creates a coin-flip when two named units share `(unitSI, unitDimension)` (Hz vs Bq). Value-based equality is the openPMD-compliant fix without those costs.

### Newton (`N`) not added

`N` (force, `unitSI=1`, dim `(1,1,-2,0,0,0,0)`) was considered for `NAMED_UNITS`. It does *not* conflict with `eV/m` / `MeV/m` as previously thought — those have `unitSI = n * e_charge`, which is not within `rel_tol=1e-9` of any SI-prefix factor times `N=1`, so the prefix fallback in `_best_atomic_match` never reaches `N` for eV-scale forces. The actual conflict is in `_best_compound_match`: `N/A` (also kg·m·A⁻¹·s⁻²) would beat `T*m` lexicographically and silently change `simplify(T*m) → N/A` — breaking the conventional magnetic-rigidity spelling. Until the compound matcher gets a bias for accelerator-physics-familiar spellings, `N` stays out and SI-magnitude forces continue to simplify to `J/m`. The comment in `NAMED_UNITS` was corrected to reflect this (the previous comment incorrectly blamed `eV/m`).

## 14. API rounding-out: `__pow__`, `compatible_with`, `conversion_factor_to`

Three small additions complete the operator/dunder set and replace ad-hoc dimension checks scattered in the package with named methods.

### `__pow__`

`pmd_unit` now supports the `**` operator, delegating to the existing `power_unit` (so the §11 compound-distribution semantics carry over for free):

```python
>>> pmd_unit("m") ** 2
pmd_unit('m^2', 1, (2.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0))
>>> pmd_unit("eV*s") ** 2                  # distributes across the compound
pmd_unit('eV^2*s^2', 2.567...e-38, (4.0, 2.0, -2.0, 0.0, 0.0, 0.0, 0.0))
>>> pmd_unit("m^2*s^2") ** 0.5
pmd_unit('m*s', 1.0, (1.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0))
```

The table entry for `norm_emit_4d` switches from `pmd_unit("m") * pmd_unit("m")` (stored symbol `"m*m"`) to `pmd_unit("m") ** 2` (stored symbol `"m^2"`). Equality is unaffected by §13, but plot labels rendered through `to_tex()` are now `\mathrm{m}^{2}` instead of `\mathrm{m}{\cdot}\mathrm{m}`.

### `compatible_with(other)`

Pure-dimension check, separate from `==`. Two units are compatible iff they share `unitDimension`; SI scale is ignored. This is what the user usually wants when asking "are these the same kind of quantity?":

```python
>>> pmd_unit("eV").compatible_with(pmd_unit("J"))     # same dim, different SI
True
>>> pmd_unit("Hz").compatible_with(pmd_unit("1/s"))   # same dim, same SI
True
>>> pmd_unit("m").compatible_with(pmd_unit("s"))
False
```

The relationship to `==` is `a == b → a.compatible_with(b)` but not the converse: `eV == J` is `False`, while `eV.compatible_with(J)` is `True`. Non-`pmd_unit` arguments raise `TypeError`.

### `conversion_factor_to(other)`

Returns the multiplicative factor mapping values from `self` to `other`. Equivalent to `self.unitSI / other.unitSI`, but guarded by a `compatible_with` check that raises `ValueError` on dimension mismatch:

```python
>>> pmd_unit("eV").conversion_factor_to(pmd_unit("J"))
1.602176634e-19
>>> pmd_unit("km").conversion_factor_to(pmd_unit("m"))
1000.0
>>> pmd_unit("Hz").conversion_factor_to(pmd_unit("1/s"))
1.0
>>> pmd_unit("m").conversion_factor_to(pmd_unit("s"))
ValueError: Incompatible units: 'm' (dimension (1.0, 0.0, ...)) cannot be converted to 's' (dimension (0.0, 0.0, 1.0, ...)).
```

### Call-site cleanup

`read_dataset_and_unit_h5` previously did:

```python
du = u / expected_unit
assert du.unitDimension == (0, 0, 0, 0, 0, 0, 0), "incompatible units"
if convert:
    fac = du.unitSI
    ...
```

The `assert` (a) is stripped under `python -O`, (b) gives a useless error message, and (c) constructs a throwaway `pmd_unit` just to read its two fields. It now reads:

```python
if not u.compatible_with(expected_unit):
    raise ValueError("Incompatible units: ...")
if convert:
    fac = u.conversion_factor_to(expected_unit)
    ...
```

No new behavior for callers passing matching units; mismatched units now raise `ValueError` with a descriptive message instead of `AssertionError` (or silently passing under `-O`).

## 15. Removed redundant `vel/c` named unit

`NAMED_UNITS` carried both `c` and `vel/c`, defined identically — dimension *velocity* `(1, 0, -1, 0, 0, 0, 0)`, `unitSI = c_light`:

```python
pmd_unit("c", c_light, "velocity"),    # speed of light
pmd_unit("vel/c", c_light, "velocity"),
```

Under the value-based equality of §13 these are the *same* unit (`pmd_unit("vel/c") == pmd_unit("c")`), so `vel/c` was a duplicate. Its symbol is also misleading: `vel/c` reads as relativistic β = v/c, but β is **dimensionless** — and is already correctly represented as `pmd_unit("1")`, which is what `pg_units` uses for `beta`, `beta_x`, `beta_y`, `beta_z`. As a velocity-dimensioned clone of `c`, `vel/c` was never actually used for β; its only consumer was the legacy `c_light` alias key. Finally, the symbol can't round-trip on its own (`vel` is not a known unit), and it was the one compound named-unit operand that `_best_atomic_match` could prefix into a non-round-tripping form such as `kvel/c`.

Fix: drop `vel/c` from `NAMED_UNITS` and repoint the legacy alias at `c`:

```diff
- pmd_unit("c", c_light, "velocity"),  # Speed of light
- pmd_unit("vel/c", c_light, "velocity"),
+ pmd_unit("c", c_light, "velocity"),  # Speed of light
...
- "c_light": pmd_unit("vel/c", c_light, "velocity"),
+ "c_light": known_unit["c"],
```

`pmd_unit("c_light")` still resolves (→ symbol `c_light`, `unitSI = c_light`), so the alias keeps working. Any HDF5 file that stored the literal `"vel/c"` symbol still reads correctly, since `read_unit_h5` reconstructs from `unitSI` + `unitDimension` rather than re-parsing the symbol string.

## 16. `divide_units` emits round-trip-safe symbols for compound divisors

§4a fixed the compound-divisor regrouping problem in `simplify()`'s *output*, but unit arithmetic itself could still produce a lying symbol: under the left-associative parser, `pmd_unit("V/m") / pmd_unit("eV/c")` emitted `"V/m/eV/c"`, which re-parses as `((V/m)/eV)/c` — a different dimension than the stored (correct) one. The in-memory object was always right; only the symbol string failed to round-trip.

Fix: a compound divisor is wrapped in parentheses, which the parser now understands as grouping (§18):

```python
>>> (pmd_unit("V/m") / pmd_unit("eV/c")).unitSymbol   # was "V/m/eV/c"
'V/m/(eV/c)'                                          # round-trips
```

A compound *numerator* needs no such handling: appended operators apply to the whole accumulated value under left association, so `a*b/c...` chains were already correct. An empty numerator symbol (the dimensionless identity) is now emitted as `"1"` (`"1/m"`, not the malformed `"/m"` that §10's parser hardening would reject).

(An earlier revision of this PR distributed the division across the divisor's tokens instead, emitting `"V/m/eV*c"` — correct but hard to read. Once the parser learned parenthesized grouping, the wrapped form became both simpler and clearer.)

## 17. Unicode look-alike characters accepted on parse

The §9 ohm uses U+03A9 GREEK CAPITAL OMEGA (the codepoint Unicode itself recommends — NFC normalizes U+2126 OHM SIGN *to* U+03A9), and `SHORT_PREFIX_FACTOR` keys micro on U+00B5 MICRO SIGN. But users paste whichever variant their keyboard/PDF produces, and the visually identical sibling codepoints were rejected with `Unknown unitSymbol`.

`from_symbol` now folds homoglyphs before lookup: NFC normalization (which maps OHM SIGN → GREEK OMEGA) plus an explicit GREEK SMALL MU (U+03BC) → MICRO SIGN (U+00B5) translation:

```python
>>> pmd_unit("Ω") == pmd_unit("Ω")   # OHM SIGN == GREEK OMEGA
True
>>> pmd_unit("μm").unitSI                  # Greek mu parses as micro
1e-06
```

The stored symbol is the folded form, so either spelling hashes and compares identically. The direct-construction path (`unitSI != 0`) stores symbols verbatim, consistent with how whitespace canonicalization (§10) is also parse-only.

## 18. Parenthesized grouping in the symbol grammar

The parser was flat and left-associative with no grouping: `"V/eV/c"` could only mean `(V/eV)/c`, and `V/(eV/c)` was inexpressible. That single limitation was the structural root cause of a string of workarounds in this PR — `power_unit`'s exponent distribution (§11), `simplify`'s atomic-operand restriction (§4a/§11), and `divide_units`' divisor handling (§16) all exist because a sub-expression couldn't be written as one operand.

The grammar now supports parentheses:

- **`(expr)` as an operand.** `_parse_single_unit` recognizes a balanced parenthesized token and recursively parses the inner expression via `from_symbol`. The tokenizer already refused to split inside parens, so `"V/(eV/c)"` tokenizes as `["V", "(eV/c)"]` and the group divides as a unit:

  ```python
  >>> pmd_unit("V/(eV/c)").unitDimension     # V / (eV/c)
  (1.0, 0.0, -2.0, -1.0, 0.0, 0.0, 0.0)
  >>> pmd_unit("V/eV/c").unitDimension       # (V/eV)/c — different!
  (-1.0, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0)
  ```

- **`(expr)^N` and `(expr)^(n/d)`.** The existing power branches recurse on their base, which now lands in the paren branch — so group exponentiation works with no extra code: `pmd_unit("(eV*s)^2") == pmd_unit("eV*s") ** 2`, and `"(m^2*s^2)^(1/2)"` is `m*s`.

- **Canonicalization.** Whitespace inside a group is normalized (`"V / ( eV / c )"` stores as `"V/(eV/c)"`), redundant parens around an atomic token are dropped (`"V/(m)"` → `"V/m"`), and nesting works (`"W/(V/(eV/c))"`).

- **Malformed input rejected.** `"()"`, `"V/()"`, `"(m"`, `"m)"`, `"(a)(b)"`, and `"(m*)"` all raise `ValueError`. A balanced-parens guard keeps adjacent groups from mis-matching as one group.

- **`to_tex` renders groups.** `pmd_unit("V/(eV/c)").to_tex()` → `\mathrm{V}/(\mathrm{eV}/\mathrm{c})`.

Old files are unaffected in both directions: symbols written by earlier versions contain no parens, and `read_unit_h5` reconstructs units from `unitSI` + `unitDimension` rather than re-parsing symbols. The §11 distribution and §4a restrictions are kept as-is (distribution gives the more conventional `eV^2*s^2` spelling for powers; `simplify` output stays atomic-only for readability) — parens simply make the grammar able to express grouping where it matters, and `divide_units` (§16) now uses them.

## 19. Package-wide unit-usage sweep

A review of every consumer of `beamphysics.units` across the package (particles, statistics, standards, plot, labels, readers, writers, fieldmesh, interfaces, wakefields). Fixes, in rough order of severity:

### Bugs fixed

- **Float-invariant hole in named units** (`units.py`): the constructor's *string*-dimension path returned the raw int tuples from `DIMENSION`, so every `NAMED_UNITS` entry carried int dims and an int `unitSI` — and `write_unit_h5` emitted **int64** HDF5 attrs where openPMD specifies float64. `dimension()` now coerces through `make_dimension`, `unitSI` is coerced to `float`, and `write_unit_h5` writes explicit `float64` attrs. A regression test asserts every construction path is float-typed.
- **`sqrt(m)*sqrt(m)` failed to parse** (`units.py`): the whole-string `sqrt(...)` match consumed the product as one sqrt with inner `m)*sqrt(m`. Both sqrt branches now require balanced parens in the inner expression (same guard as the paren-group branch). Found because covariance units like `cov_x_bar__x_bar` produce exactly this symbol.
- **Plot labels glued SI prefixes onto compound symbols** (`plot.py`): `prefix + symbol` is wrong for powered/sqrt/reciprocal symbols (`'m'+'m^2'` reads as `(mm)² = 1e-6` while the data scale is 1e-3 — including a label introduced in this PR by the `norm_emit_4d` `m*m → m^2` change; `'m'+'sqrt(m)'` is unparseable). The scaling-to-symbol logic is the single method `pmd_unit.scaled_symbol(factor)`: the given spelling is kept when an SI prefix can express the factor on it (`meV/c`, `MV/m`), the scaled value re-spells via `simplify` when it cannot (`1e-6 * mA*s` → `nC`, `1e3 * 1/s` → `kHz`), and the factor is written out otherwise (`1e-3 m^2`). Callers pass the scale factor that `nice_array`/`nice_scale_prefix`/`plottable_array` already return — no prefix-string round trips. Unparseable user unit strings are handled only at the one boundary where they arrive, `plottable_array_and_units` (degrading to `'1e3 counts'`). `plot.py` keeps only the plot-domain `_charge_density_units_str` policy; the units→TeX rendering lives inside `mathlabel` (which, with no keys, returns a units-only label). A combined `units.plottable_array_and_units(x, units, ...)` returns the scaled array together with its display-units string, so no plot call site handles a bare SI prefix or factor-to-prefix conversion anymore (~10 two-step sites collapsed, including two missed hand-glued labels in `wakefield_plot`). `labels.texlabel`'s never-taken `None` return was removed from its docstring along with `mathlabel`'s dead `or`-fallback. The marginal/density histogram labels keep each axis's own prefix in the denominator, as before — `nC/µm` against an x axis in µm, `pC/(keV/c)` against keV/c (now with correct grouping; the bare `C/keV/c` would misread) — with the time axis still special-cased to amps. The CDF label no longer double-prefixes (`1e-6·mA·s` now labels `µC`, not `µmC`).
- **HDF5 unitSymbol bytes/str mismatch** (`writers.py` + `units.py`): `write_component_data` wrote `np.bytes_` symbols that `read_unit_h5` never decoded, so units read back from package-written files had bytes symbols (`u * pmd_unit('s')` raised `TypeError`). `write_component_data` now delegates to `write_unit_h5` (UTF-8 str, required for `Ω`/`µ`), and `read_unit_h5` decodes bytes from older files.
- **`is_constant_component` precedence** (`readers.py`): `"value" and "shape" in h5.attrs` only checked `shape`.
- **`nice_scale_prefix` leaked `np.float64`** into its public return; now a plain `float`.

### Simplifications / cleanups

- `slice_plot` / `density_and_slice_plot` check key-unit consistency by comparing `pmd_unit` objects (value-based) instead of symbol strings, and raise `ValueError` instead of a bare `assert`.
- `parse_bunching_str` raises `ValueError` instead of `assert`.
- `pg_units` gained the slice-statistics keys `current` (A) and `density` (C/m).
- `readers.component_str` describes unknown dimensions instead of raising `KeyError`.
- The fieldmesh rectangular plot labels render units via `to_tex()` / safe prefixing (matching the cylindrical plot), and the `Re[...]`/`Im[...]` legend brackets are balanced.
- Stale doctests in `units.py` fixed (including a long-wrong `eV/c` `unitSI` value), and a test now runs the module's doctests so they stay honest.
- Maintainability pass: removed the never-used long-form prefix tables (`PREFIX_FACTOR`/`PREFIX`), a dead constructor bootstrap clause (`or not known_unit` — every import-time unit passes an explicit `unitSI`), and the redundant `__ne__` (Python derives `!=` from `__eq__`). `texlabel` gained the missing `ptp_` formatter (it was listed as an operator prefix but silently fell through to the raw fallback). `mathlabel` with no keys now returns a units-only label (used by the marginal histograms), so the interim `units_tex`/`units_label` helpers folded back into it — `labels.py` exposes exactly two functions, `texlabel` and `mathlabel`.

### Compatibility notes

- `unitDimension`/`unitSI` HDF5 attrs are now always written as float64 (spec-conformant; previously int64 for named units). `unitSymbol` is written as a UTF-8 str attr rather than fixed-length bytes; h5py and openPMD tooling read both, and `read_unit_h5` accepts both.

### Split into a follow-up PR

The same review found several non-unit bugs that are deliberately **not** in this PR, to keep it units-focused: the Genesis2 `.dpa` momentum reconstruction typo (`py * 2` → `py**2`), twiss ignoring particle weights, the statistics standard documenting dimensionally wrong covariance units (`eV/c^2`) and missing `y_bar`/`py_bar` keys, fieldmesh file validation via strippable `assert`s, a dead `units` dict in `genesis4_par_to_data`, and the inverted `average_current` docstring formula.

## 20. Single recursive parser

`from_symbol` and `_parse_single_unit` had grown into near-duplicates: both did the known-unit copy lookup and the sqrt branch, while compound tokenization lived only in the former and power notation/the SI-prefix fallback only in the latter. The duplication had already drifted into a bug: `from_symbol`'s sqrt branch recursed into `from_symbol` (compound-capable), but `_parse_single_unit`'s recursed into *itself* — so `sqrt(kg*m)` parsed while the identical token inside `sqrt(kg*m)*s` raised `Unknown unitSymbol: kg*m`.

`_parse_single_unit` is gone. `from_symbol` is now the single recursive parser:

1. normalize (NFC + homoglyph fold + whitespace, including around `^`);
2. known-unit/alias lookup;
3. tokenize on top-level `*`/`/`; multiple tokens recurse through `from_symbol` and combine;
4. single tokens fall through sqrt → paren group → power (fraction and plain) → SI-prefix fallback, each recursing through `from_symbol` on a strictly shorter substring (so parsing terminates).

Every notation now works uniformly at any recursion depth; `sqrt(kg*m)*s`, `V/sqrt(kg*m)`, and `sqrt(kg*m)^2` all parse and round-trip. Single-token spellings are still preserved in the stored symbol (`m^(1/2)`, `(eV/c)^2`), with redundant parens dropped (`(m)` → `m`).

## Files changed

- `beamphysics/units.py` — fixes for §2, §4–§18 (including `to_tex`, value-based `__eq__`/`__hash__`, `power_unit` distribution, input hardening, sqrt canonicalization, `__pow__` / `compatible_with` / `conversion_factor_to`, the `read_dataset_and_unit_h5` call-site cleanup, round-trip-safe `divide_units`, Unicode homoglyph folding, and parenthesized grouping in the parser).
- `beamphysics/labels.py` — `mathlabel` now uses `pmd_unit.to_tex()` to render unit labels, with the unparseable-string and dimensionless edge cases handled (§12).
- `tests/test_unit.py` — beamphysics-side regression tests for §2–§14, §16–§18.
- `tests/test_labels.py` — new; `mathlabel` rendering tests for §12 (both TeX and non-TeX branches, including the dimensionless and unparseable-string edge cases).
- `tests/test_units_pint.py` — new, independent `pint` cross-check suite (also drops the now-dead `vel/c` entry from its symbol→pint table, §15).
- `docs/examples/units.ipynb` — documents power notation, SI prefixes, the improved `simplify`, canonical-symbol normalization, malformed-input rejection, `to_tex()` + `mathlabel`, value-based equality, and the new `**` / `compatible_with` / `conversion_factor_to` API (§14). Also adds a parenthesized-grouping section (§18) with canonicalization examples, and includes a grouped symbol in the `to_tex()` demo. Cell count 84 → 76 after consolidating overlapping power-notation demos.
- `beamphysics/interfaces/genesis.py` — switch three call sites from the deprecated `unit` alias to `pmd_unit`.
- `beamphysics/plot.py` — labels via `plottable_array_and_units` / `pmd_unit.scaled_symbol`; corrected density/CDF/fieldmesh labels (§19).
- `beamphysics/writers.py`, `beamphysics/readers.py` — unit attrs via `write_unit_h5`; bytes decode; `is_constant_component` fix; graceful unknown dimensions (§19).
- `tests/test_plot_labels.py` — new; plot label tests (scaled_symbol policy, charge-density labels, marginal axis prefixes) and single-particle draw test (§19).
- `pyproject.toml`, `environment.yml` — add `pint` as a test-only dependency.
- `.gitignore` — ignore generated `tests/artifacts/`.

## Test results

```
1796 passed, 1 xpassed, 3 warnings in ~10s
```

(The single xpass is `test_wavefront.py::test_gaussian_plot_fluence[logscale]`, a pre-existing environment-dependent xfail unrelated to this PR.)

Key new tests:

- `test_simplify_examples` — the §3 examples plus `J/C → V`.
- `test_simplify_electric_potential_volt` — locks in the §2 dimension.
- `test_simplify_division_by_compound_round_trips` — §4a regression.
- `test_simplify_dimensionless_never_bare_prefix` — §4b regression.
- `test_simplify_never_breaks_round_trip` — invariant over all named-unit products/quotients.
- `test_parsing_compound_does_not_mutate_named_units` — §5 regression.
- `test_si_prefix_parsing`, `test_bare_prefix_and_junk_rejected`, `test_prefix_does_not_shadow_base_units`, `test_simplify_prefixed_output_round_trips` — §6.
- `test_sqrt_unit_halves_dimension`, `test_sqrt_unit_symbol_round_trips`, `test_sqrt_symbol_whitespace_normalized`, `test_latex_sqrt_form_not_accepted` — §7, §10.
- `test_cov_units_with_prefixed_subkey` — §8 regression.
- `test_ohm_unit_and_alias`, `test_wakefield_unit_strings_parse` — §9.
- `test_whitespace_around_caret_tolerated`, `test_symbol_whitespace_normalized_in_stored_form`, `test_parser_rejects_empty_operands`, `test_negative_unitSI_rejected` — §10.
- `test_power_unit_distributes_over_compound`, `test_simplify_skips_compound_numerator` — §11.
- `test_to_tex_translation`, `test_to_tex_preserves_input_structure` — §12.
- `test_equality_is_value_based_per_standard`, `test_hash_distinguishes_same_dimension_units` — §13.
- `test_pow_operator_matches_power_unit`, `test_compatible_with_is_dimension_only`, `test_conversion_factor_to_round_trips_and_rejects_mismatch`, `test_norm_emit_4d_stored_as_m_squared` — §14.
- `test_divide_by_compound_divisor_round_trips`, `test_divide_with_empty_numerator_symbol` — §16.
- `test_unicode_homoglyphs_normalized` — §17.
- `test_parenthesized_grouping_parses`, `test_parenthesized_group_with_power`, `test_malformed_parens_rejected`, `test_divide_units_emits_parenthesized_divisor` — §18 (plus parenthesized/nested symbols added to the pint cross-check's compound-symbol table).
- `tests/test_labels.py` (new) — `mathlabel` rendering through `to_tex()`, the `{\cdot}` fallback, and the dimensionless non-TeX edge case — §12.

### Independent `pint` cross-check (`tests/test_units_pint.py`)

`pint` is used as an external oracle (test-only; the package never imports it). It verifies, against `pint`'s own dimensional analysis:

- the `DIMENSION` dict (every named dimension) — this is the check that would have caught the §2 `electric_potential` bug;
- every `NAMED_UNITS` entry (dimension **and** SI scale);
- compound-symbol parsing, SI-prefixed symbols, and fractional/`sqrt` powers (e.g. spectral densities like `V/m/Hz^0.5`);
- `simplify()` preserves physical meaning;
- the entire `pg_units` table (all keys, plus operator/`cov_`/field-prefix rules), with a guard that every key has an independent expectation;
- impedance / wakefield units (`Ω`, `Ω/m`, `V/A`, `V/C`, `V/C/m`).

## Acknowledgment

The review, bug fixes (§2, §4–§9), the independent `pint` cross-check suite, the regression tests, and the updated `docs/examples/units.ipynb` were developed in collaboration with **Claude** (Anthropic's Claude Opus 4.8, via Claude Code). It traced the `electric_potential` dimension error, the `simplify` round-trip and `NAMED_UNITS` mutation bugs, the `sqrt_unit` floor-division bug, and the `cov_` key-parsing bug, and proposed the `pint` oracle that anchors the units to an independent reference. A follow-up review pass (Claude Fable 5) found and fixed the two `mathlabel` edge cases (§12), the `divide_units` compound-divisor symbol (§16), and the Unicode homoglyph rejection (§17), and added parenthesized grouping to the symbol grammar (§18).

### DIFF
--- a/beamphysics/units.py
+++ b/beamphysics/units.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import copy
 import math
 import re
+import unicodedata
 import warnings
 from typing import Optional, Sequence
 
@@ -32,6 +33,11 @@ Limit = tuple[Optional[float], Optional[float]]
 # fractional powers (e.g. from sqrt_unit) round-trip through the same code
 # path as integer ones. Equality with int tuples still works (``1.0 == 1``).
 Dimension = tuple[float, float, float, float, float, float, float]
+
+# Homoglyph folding applied to symbols on parse (after NFC normalization,
+# which already folds U+2126 OHM SIGN into U+03A9 GREEK CAPITAL OMEGA):
+# GREEK SMALL MU -> MICRO SIGN, the codepoint SHORT_PREFIX_FACTOR uses.
+_HOMOGLYPH_TRANSLATION = str.maketrans({"μ": "µ"})
 
 # Module-level dict that will be populated after NAMED_UNITS is created
 # This avoids the globals() check chicken-and-egg problem
@@ -190,7 +196,17 @@ class pmd_unit:
         # stripped. We do not split on whitespace (no implicit space-as-*),
         # which would conflict with multi-token named units like "charge #"
         # and with the SI-prefix fallback ("k eV" vs "keV").
-        unitSymbol = unitSymbol.strip()
+        #
+        # Unicode homoglyphs are folded to the codepoints used by the unit
+        # tables: NFC maps U+2126 OHM SIGN to U+03A9 GREEK CAPITAL OMEGA
+        # (the codepoint Unicode itself recommends, and the one NAMED_UNITS
+        # uses), and GREEK SMALL MU (U+03BC) is translated to MICRO SIGN
+        # (U+00B5), which is what SHORT_PREFIX_FACTOR keys on.
+        unitSymbol = (
+            unicodedata.normalize("NFC", unitSymbol)
+            .translate(_HOMOGLYPH_TRANSLATION)
+            .strip()
+        )
 
         # Check if known_unit dict has been populated
         if not known_unit:
@@ -877,7 +893,19 @@ def divide_units(u1: pmd_unit, u2: pmd_unit) -> pmd_unit:
     if s1 == s2:
         symbol = "1"
     else:
-        symbol = s1 + "/" + s2
+        # The parser is flat and left-associative ("a/b/c" means (a/b)/c), so
+        # writing f"{s1}/{s2}" with a *compound* divisor would regroup:
+        # V/(eV/c) emitted as "V/eV/c" reads back as V/(eV*c). Distribute the
+        # division across the divisor's tokens instead, inverting each of its
+        # operators: a/(b*c) -> a/b/c and a/(b/c) -> a/b*c. A compound
+        # numerator needs no such care — appended operators apply to the whole
+        # accumulated value under left association. An empty numerator symbol
+        # (the dimensionless identity) is emitted as "1" so the result does
+        # not start with a bare operator.
+        numerator = s1 if s1 else "1"
+        parts, ops = _tokenize_compound(s2)
+        inverted = ["/"] + ["*" if op == "/" else "/" for op in ops]
+        symbol = numerator + "".join(op + p for op, p in zip(inverted, parts))
     d1 = u1.unitDimension
     d2 = u2.unitDimension
     dim = tuple(a - b for a, b in zip(d1, d2))
--- a/beamphysics/units.py
+++ b/beamphysics/units.py
@@ -360,6 +360,81 @@ class pmd_unit:
     def __truediv__(self, other) -> pmd_unit:
         return divide_units(self, other)
 
+    def __pow__(self, power: float) -> pmd_unit:
+        return power_unit(self, power)
+
+    def compatible_with(self, other: pmd_unit) -> bool:
+        """Return ``True`` when ``other`` has the same dimension as ``self``.
+
+        Compatibility is a pure-dimension check: two units are compatible
+        iff a value expressed in one can be converted to the other by a
+        single multiplicative SI factor (see :meth:`conversion_factor_to`).
+        It does *not* require equal :attr:`unitSI`, so e.g. ``eV`` and
+        ``J`` are compatible.
+
+        Parameters
+        ----------
+        other : pmd_unit
+            The unit to compare against.
+
+        Returns
+        -------
+        bool
+            ``True`` iff ``self.unitDimension == other.unitDimension``.
+
+        Examples
+        --------
+        >>> pmd_unit("eV").compatible_with(pmd_unit("J"))
+        True
+        >>> pmd_unit("Hz").compatible_with(pmd_unit("1/s"))
+        True
+        >>> pmd_unit("m").compatible_with(pmd_unit("s"))
+        False
+        """
+        if not isinstance(other, pmd_unit):
+            raise TypeError(
+                f"compatible_with expects a pmd_unit, got {type(other).__name__}"
+            )
+        return self.unitDimension == other.unitDimension
+
+    def conversion_factor_to(self, other: pmd_unit) -> float:
+        """Return the factor that converts a value in ``self`` to ``other``.
+
+        For a value ``x`` measured in ``self``, the value in ``other`` is
+        ``x * self.conversion_factor_to(other)``. Equivalent to
+        ``self.unitSI / other.unitSI`` after a dimension check.
+
+        Parameters
+        ----------
+        other : pmd_unit
+            The target unit. Must have the same :attr:`unitDimension` as
+            ``self``.
+
+        Returns
+        -------
+        float
+            Multiplicative factor mapping values in ``self`` to ``other``.
+
+        Raises
+        ------
+        ValueError
+            If ``other`` is not dimensionally :meth:`compatible_with` ``self``.
+
+        Examples
+        --------
+        >>> pmd_unit("eV").conversion_factor_to(pmd_unit("J"))
+        1.602176634e-19
+        >>> pmd_unit("km").conversion_factor_to(pmd_unit("m"))
+        1000.0
+        """
+        if not self.compatible_with(other):
+            raise ValueError(
+                f"Incompatible units: {self.unitSymbol!r} (dimension "
+                f"{self.unitDimension}) cannot be converted to "
+                f"{other.unitSymbol!r} (dimension {other.unitDimension})."
+            )
+        return self.unitSI / other.unitSI
+
     def __eq__(self, other) -> bool:
         # Per the openPMD standard, a unit is defined by (unitSI,
         # unitDimension); the symbol is a presentational hint. Two units are
@@ -1263,7 +1338,7 @@ for k in ["power"]:
 for k in ["norm_emit_x", "norm_emit_y"]:
     PARTICLEGROUP_UNITS[k] = pmd_unit("m")
 for k in ["norm_emit_4d"]:
-    PARTICLEGROUP_UNITS[k] = pmd_unit("m") * pmd_unit("m")
+    PARTICLEGROUP_UNITS[k] = pmd_unit("m") ** 2
 for k in ["Lz"]:
     PARTICLEGROUP_UNITS[k] = pmd_unit("m") * pmd_unit("eV/c")
 for k in ["xp", "yp"]:
@@ -1422,13 +1497,16 @@ def read_dataset_and_unit_h5(h5, expected_unit=None, convert=True):
         # Try to get unit
         expected_unit = pmd_unit(expected_unit)
 
-    # Check dimensions
-    du = u / expected_unit
-
-    assert du.unitDimension == (0, 0, 0, 0, 0, 0, 0), "incompatible units"
+    if not u.compatible_with(expected_unit):
+        raise ValueError(
+            f"Incompatible units: file unit {u.unitSymbol!r} (dimension "
+            f"{u.unitDimension}) does not match expected "
+            f"{expected_unit.unitSymbol!r} (dimension "
+            f"{expected_unit.unitDimension})."
+        )
 
     if convert:
-        fac = du.unitSI
+        fac = u.conversion_factor_to(expected_unit)
         return fac * np.array(h5), expected_unit
     else:
         return np.array(h5), u
--- a/beamphysics/units.py
+++ b/beamphysics/units.py
@@ -168,6 +168,13 @@ class pmd_unit:
         ValueError
             If the symbol is not found in known_unit dict or cannot be parsed.
         """
+        # Strip outer whitespace so users can write "  eV/c" or "kg*m / s".
+        # Per-token whitespace inside the operator-split branch below is also
+        # stripped. We do not split on whitespace (no implicit space-as-*),
+        # which would conflict with multi-token named units like "charge #"
+        # and with the SI-prefix fallback ("k eV" vs "keV").
+        unitSymbol = unitSymbol.strip()
+
         # Check if known_unit dict has been populated
         if not known_unit:
             raise ValueError(
@@ -219,6 +226,7 @@ class pmd_unit:
                     current += char
 
             parts.append(current)  # Don't forget the last part
+            parts = [p.strip() for p in parts]
 
             # Parse the first part (may include ^power)
             result = cls._parse_single_unit(parts[0])
@@ -576,10 +584,12 @@ def multiply_units(u1: pmd_unit, u2: pmd_unit) -> pmd_unit:
 
     s1 = u1.unitSymbol
     s2 = u2.unitSymbol
-    if s1 == s2:
-        symbol = f"({s1})^2"
-    else:
-        symbol = s1 + "*" + s2
+    # No special case for s1 == s2: writing "s1^2" would be wrong when s1 is
+    # itself compound (e.g. "kg*m^2" parses as kg*(m^2), not (kg*m)^2), and
+    # wrapping in parens ("(s1)^2") produces a symbol the parser cannot read
+    # back. The plain "s1*s2" form always round-trips and yields the correct
+    # dimension regardless of whether either operand is atomic or compound.
+    symbol = s1 + "*" + s2
     d1 = u1.unitDimension
     d2 = u2.unitDimension
     dim = tuple(sum(x) for x in zip(d1, d2))
@@ -721,9 +731,11 @@ def make_dimension(dim: Sequence[int | float]) -> Dimension:
     The openPMD standard defines ``unitDimension`` as an array of 7
     ``float64`` values, so floats are used uniformly here. This also lets
     fractional exponents (e.g. from :func:`sqrt_unit`) share a single
-    representation with integer ones.
+    representation with integer ones. Negative zeros (from e.g. ``-1.0 * 0.0``
+    inside :func:`power_unit`) are normalized to positive ``0.0`` so the
+    ``repr`` is stable and exponent tuples compare cleanly as dict keys.
     """
-    dim = tuple(float(d) for d in dim)
+    dim = tuple(0.0 if d == 0 else float(d) for d in dim)
     if len(dim) != 7:
         raise ValueError("Dimensions must be 7 elements.")
     return dim
@@ -787,6 +799,11 @@ NAMED_UNITS = [
     pmd_unit("eV", e_charge, "energy"),
     pmd_unit("J", 1, "energy"),
     pmd_unit("eV/c", e_charge / c_light, "momentum"),
+    # The Newton as the force (dimension L M T^-2) is intentionally represented as eV/m and J/m
+    # rather than N: the accelerator-physics convention writes accelerating
+    # gradients as eV/m or MV/m, and adding "N" here would beat them in
+    # _symbol_complexity (shorter symbol wins) and change existing simplify()
+    # output.
     pmd_unit("eV/m", e_charge, (1, 1, -2, 0, 0, 0, 0)),
     pmd_unit("J/m", 1, (1, 1, -2, 0, 0, 0, 0)),
     pmd_unit("W", 1, (2, 1, -3, 0, 0, 0, 0)),
--- a/beamphysics/units.py
+++ b/beamphysics/units.py
@@ -112,11 +112,17 @@ class pmd_unit:
     >>> pmd_unit.from_symbol('A*s')
     pmd_unit('A*s', 1, (0.0, 0.0, 1.0, 1.0, 0.0, 0.0, 0.0))
 
-    Simple equalities are possible:
+    Equality follows the openPMD standard: the symbol is presentational and
+    two units are equal when they share the same ``unitDimension`` and
+    ``unitSI`` (compared with a small relative tolerance):
 
     >>> pmd_unit("T") == pmd_unit("T")
     True
     >>> pmd_unit("eV") == pmd_unit("T")
+    False
+    >>> pmd_unit("1/s") == pmd_unit("Hz")
+    True
+    >>> pmd_unit("J/m") == pmd_unit("eV/m")  # different SI scales, unequal
     False
     """
 
@@ -126,6 +132,17 @@ class pmd_unit:
         unitSI: int | float = 0,
         unitDimension: str | Sequence[int | float] = (0, 0, 0, 0, 0, 0, 0),
     ):
+        # The openPMD standard defines unitSI as a positive conversion factor
+        # to the corresponding SI unit. Negative scales would invert the sign
+        # of every value an instrument writes through this unit, so they are
+        # rejected here. (unitSI=0 is permitted as the sentinel that triggers
+        # symbol lookup below.)
+        if unitSI < 0:
+            raise ValueError(
+                f"unitSI must be non-negative (got {unitSI!r}). "
+                "The openPMD standard defines unitSI as a conversion factor to SI."
+            )
+
         # Use the provided values directly when unitSI is given explicitly, or
         # while NAMED_UNITS is still being built (known_unit not yet populated).
         # Otherwise look the symbol up / parse it.
@@ -192,41 +209,29 @@ class pmd_unit:
             result._unitSymbol = unitSymbol
             return result
 
-        # Handle sqrt() notation: sqrt(X) = X^0.5
-        sqrt_match = re.match(r"^sqrt\((.+)\)$", unitSymbol)
+        # Handle sqrt() notation: sqrt(X) = X^0.5. The stored symbol is
+        # always the canonical, whitespace-free "sqrt(<inner>)" form so that
+        # "sqrt( m )", "sqrt(m)", and "  sqrt(m)" all compare equal and hash
+        # identically.
+        sqrt_match = re.match(r"^sqrt\(\s*(.+?)\s*\)$", unitSymbol)
         if sqrt_match:
             inner = sqrt_match.group(1)
             base_unit = cls.from_symbol(inner)
             result = power_unit(base_unit, 0.5)
-            # Preserve the original sqrt notation in the symbol
-            result._unitSymbol = unitSymbol
+            result._unitSymbol = f"sqrt({base_unit.unitSymbol})"
             return result
 
         # Not in known_unit - try operator parsing (e.g., "A*s", "kg*m/s", "m^2")
         if "*" in unitSymbol or "/" in unitSymbol or "^" in unitSymbol:
-            # Tokenize: split by * and / but not inside parentheses
-            # This allows m^(-3/2) to remain intact
-            parts = []
-            operators = []
-            current = ""
-            paren_depth = 0
+            parts, operators = _tokenize_compound(unitSymbol)
 
-            for char in unitSymbol:
-                if char == "(":
-                    paren_depth += 1
-                    current += char
-                elif char == ")":
-                    paren_depth -= 1
-                    current += char
-                elif char in "*/" and paren_depth == 0:
-                    parts.append(current)
-                    operators.append(char)
-                    current = ""
-                else:
-                    current += char
-
-            parts.append(current)  # Don't forget the last part
-            parts = [p.strip() for p in parts]
+            # Reject malformed compound input (empty operand around an
+            # operator). Previously these silently parsed because the empty
+            # token resolved to the identity (""): "m*" -> m*1 = m, "/m" -> 1/m.
+            if any(p == "" for p in parts):
+                raise ValueError(
+                    f"Malformed unit symbol {unitSymbol!r}: empty operand around an operator."
+                )
 
             # Parse the first part (may include ^power)
             result = cls._parse_single_unit(parts[0])
@@ -239,8 +244,10 @@ class pmd_unit:
                 elif op == "/":
                     result = result / unit_obj
 
-            # Preserve the original symbol but keep computed dimension
-            result._unitSymbol = unitSymbol
+            # Canonical symbol: stripped tokens joined by the original
+            # operators with no whitespace. So "eV / c" and "eV/c" share a
+            # symbol (and therefore compare equal / hash identically).
+            result._unitSymbol = _join_compound(parts, operators)
             return result
 
         # Single token, not directly known: delegate to _parse_single_unit,
@@ -264,6 +271,10 @@ class pmd_unit:
         pmd_unit
             The parsed unit object.
         """
+        # Strip whitespace around the token and around any "^" so callers can
+        # write "m ^ 2" or "  m^2" interchangeably with "m^2".
+        part = re.sub(r"\s*\^\s*", "^", part.strip())
+
         # Check if it's a known unit first. Return a shallow copy, never the
         # shared cached object: callers (e.g. from_symbol) reassign
         # ``_unitSymbol`` on the result, which would otherwise mutate the
@@ -273,13 +284,14 @@ class pmd_unit:
             result._unitSymbol = part
             return result
 
-        # Check for sqrt() notation
-        sqrt_match = re.match(r"^sqrt\((.+)\)$", part)
+        # Check for sqrt() notation. Whitespace inside the parens is stripped
+        # and the stored symbol is canonicalized to "sqrt(<inner>)".
+        sqrt_match = re.match(r"^sqrt\(\s*(.+?)\s*\)$", part)
         if sqrt_match:
             inner = sqrt_match.group(1)
             base_unit = cls._parse_single_unit(inner)
             result = power_unit(base_unit, 0.5)
-            result._unitSymbol = part
+            result._unitSymbol = f"sqrt({base_unit.unitSymbol})"
             return result
 
         # Check for power notation with parenthesized fraction: m^(-3/2)
@@ -323,7 +335,12 @@ class pmd_unit:
         raise ValueError(f"Unknown unitSymbol: {part}")
 
     def __hash__(self) -> int:
-        return hash((self.unitSymbol, self.unitSI, self.unitDimension))
+        # The openPMD standard defines a unit by (unitSI, unitDimension); the
+        # symbol is presentational. Hash on unitDimension alone so that any
+        # two equal units (per __eq__'s isclose comparison on unitSI) share a
+        # hash bucket — preserving the hash/eq invariant even when unitSI
+        # values arrived via slightly different arithmetic paths.
+        return hash(self.unitDimension)
 
     @property
     def unitSymbol(self) -> str:
@@ -344,9 +361,19 @@ class pmd_unit:
         return divide_units(self, other)
 
     def __eq__(self, other) -> bool:
-        if isinstance(other, self.__class__):
-            return self.__dict__ == other.__dict__
-        return False
+        # Per the openPMD standard, a unit is defined by (unitSI,
+        # unitDimension); the symbol is a presentational hint. Two units are
+        # equal when both fields match, with math.isclose on unitSI to
+        # absorb floating-point drift from arithmetic like
+        # (eV/c) * c -> eV (the eV value reconstructed through c_light
+        # round-tripping may differ from e_charge by 1 ULP).
+        if not isinstance(other, self.__class__):
+            return False
+        if self.unitDimension != other.unitDimension:
+            return False
+        return math.isclose(
+            self.unitSI, other.unitSI, rel_tol=_SIMPLIFY_REL_TOL, abs_tol=0.0
+        )
 
     def __ne__(self, other) -> bool:
         return not self.__eq__(other)
@@ -406,6 +433,81 @@ class pmd_unit:
 
         return self
 
+    def to_tex(self) -> str:
+        """Render the stored unit symbol as a matplotlib-mathtext string.
+
+        The translation is purely structural — no algebraic simplification
+        or reordering. Each atomic unit name is wrapped in ``\\mathrm{...}``
+        (so it renders upright), ``*`` becomes ``{\\cdot}``, ``/`` is kept
+        as ``/``, ``^N`` becomes ``^{N}``, and ``sqrt(X)`` becomes
+        ``\\sqrt{...}`` (recursing on the inner expression).
+
+        The result is intended to be embedded inside an ``$...$`` mathtext
+        block by the caller. The identity / empty symbol returns ``""``.
+
+        Returns
+        -------
+        str
+            A TeX fragment equivalent to :attr:`unitSymbol`.
+
+        Examples
+        --------
+        >>> pmd_unit("eV/c").to_tex()
+        '\\\\mathrm{eV}/\\\\mathrm{c}'
+
+        >>> pmd_unit("kg*m/s").to_tex()
+        '\\\\mathrm{kg}{\\\\cdot}\\\\mathrm{m}/\\\\mathrm{s}'
+
+        >>> pmd_unit("m^2").to_tex()
+        '\\\\mathrm{m}^{2}'
+
+        >>> pmd_unit("sqrt(m)").to_tex()
+        '\\\\sqrt{\\\\mathrm{m}}'
+        """
+        return _symbol_to_tex(self._unitSymbol)
+
+
+def _symbol_to_tex(symbol: str) -> str:
+    """Recursive worker for :meth:`pmd_unit.to_tex`."""
+    if symbol == "":
+        return ""
+    if symbol == "1":
+        return "1"
+
+    # Top-level compound (``a*b``, ``a/b``, ``sqrt(a)*b``). Tokenize first so
+    # the recursion handles nested ``sqrt(...)`` and per-token powers.
+    if "*" in symbol or "/" in symbol:
+        parts, operators = _tokenize_compound(symbol)
+        if len(parts) > 1:
+            tex_parts = [_symbol_to_tex(p) for p in parts]
+            out = tex_parts[0]
+            for op, p in zip(operators, tex_parts[1:]):
+                out += r"{\cdot}" + p if op == "*" else "/" + p
+            return out
+
+    # sqrt(X) -> \sqrt{ <recursed X> }
+    sqrt_match = re.match(r"^sqrt\((.+)\)$", symbol)
+    if sqrt_match:
+        inner = _symbol_to_tex(sqrt_match.group(1))
+        return rf"\sqrt{{{inner}}}"
+
+    # base^(n/d) -> \mathrm{base}^{n/d}
+    paren_power_match = _TOKEN_PAREN_POWER_RE.match(symbol)
+    if paren_power_match:
+        base = _symbol_to_tex(paren_power_match.group(1))
+        num = paren_power_match.group(2)
+        den = paren_power_match.group(3)
+        return rf"{base}^{{{num}/{den}}}"
+
+    # base^N -> \mathrm{base}^{N}
+    power_match = _TOKEN_POWER_RE.match(symbol)
+    if power_match:
+        base = _symbol_to_tex(power_match.group(1))
+        return f"{base}^{{{power_match.group(2)}}}"
+
+    # Atomic name (possibly with internal whitespace, e.g. "charge #").
+    return rf"\mathrm{{{symbol}}}"
+
 
 def is_dimensionless(u: pmd_unit) -> bool:
     """Checks if the unit is dimensionless"""
@@ -423,6 +525,49 @@ def is_identity(u: pmd_unit) -> bool:
 
 _SIMPLIFY_REL_TOL = 1e-9
 _DIMENSIONLESS: Dimension = (0, 0, 0, 0, 0, 0, 0)
+
+
+def _tokenize_compound(symbol: str) -> tuple[list[str], list[str]]:
+    """Split ``symbol`` on top-level ``*`` and ``/`` (respecting parentheses).
+
+    Returns ``(parts, operators)`` where each part is whitespace-stripped and
+    ``operators[i]`` joins ``parts[i]`` to ``parts[i+1]``. Used by both the
+    parser and :func:`power_unit` (for distributing an exponent across a
+    compound symbol).
+    """
+    parts: list[str] = []
+    operators: list[str] = []
+    current = ""
+    paren_depth = 0
+    for char in symbol:
+        if char == "(":
+            paren_depth += 1
+            current += char
+        elif char == ")":
+            paren_depth -= 1
+            current += char
+        elif char in "*/" and paren_depth == 0:
+            parts.append(current)
+            operators.append(char)
+            current = ""
+        else:
+            current += char
+    parts.append(current)
+    return [p.strip() for p in parts], operators
+
+
+def _join_compound(parts: Sequence[str], operators: Sequence[str]) -> str:
+    """Inverse of :func:`_tokenize_compound`: re-emit a flat compound symbol."""
+    out = parts[0]
+    for op, p in zip(operators, parts[1:]):
+        out += op + p
+    return out
+
+
+def _is_atomic_symbol(symbol: str) -> bool:
+    """True if the symbol has no top-level ``*`` or ``/`` (parens are kept)."""
+    parts, _ = _tokenize_compound(symbol)
+    return len(parts) == 1
 
 
 def _symbol_complexity(symbol: str) -> tuple[int, int, str]:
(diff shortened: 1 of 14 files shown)


beamphysics/units.py
```diff
@@ -252,9 +260,12 @@ def _parse_single_unit(cls, part: str) -> pmd_unit:
...
             The parsed unit object.
         """
-        # Check if it's a known unit first
+        # Check if it's a known unit first. Return a copy, never the shared
+        # cached object: callers (e.g. from_symbol) reassign ``_unitSymbol`` on
+        # the result, which would otherwise mutate the global NAMED_UNITS entry.
         if part in known_unit:
-            return known_unit[part]
+            u = known_unit[part]
+            return cls(part, unitSI=u.unitSI, unitDimension=u.unitDimension)
```
electronsandstuff: Use `copy.copy(...)` here?

beamphysics/units.py
```diff
@@ -366,6 +415,137 @@ def is_identity(u: pmd_unit) -> bool:
...
+def _best_atomic_match(
+    target_dim: Dimension,
+    target_si: float,
+    named_units: Sequence[pmd_unit],
+) -> pmd_unit | None:
+    """
+    Find the simplest named unit (optionally with an SI prefix) whose
+    dimension and SI value match the target.
+    """
+    candidates: list[tuple[int, tuple[int, int, str], pmd_unit, str, float]] = []
```
electronsandstuff: The objects in this list seem complicated. Use dict / named dict / dataclass instead.

beamphysics/units.py
```diff
@@ -366,6 +415,137 @@ def is_identity(u: pmd_unit) -> bool:
...
+    best: pmd_unit | None = None
+    best_key: tuple[int, int, str] | None = None
+
+    for a in (u for group in dim_index.values() for u in group):
+        a_si = a.unitSI
+
+        # self = a * b  =>  b_dim = target - a_dim, b_si = target / a
+        mul_dim = tuple(td - ad for td, ad in zip(target_dim, a.unitDimension))
+        mul_target_si = target_si / a_si
+        for b in dim_index.get(mul_dim, ()):
```
electronsandstuff, line 863: This is a O(n^2) brute-force search in python for loops. How often is this called? On unit creation?

beamphysics/units.py
```diff
@@ -470,14 +638,15 @@ def sqrt_unit(u: pmd_unit) -> pmd_unit:
     if not isinstance(u, pmd_unit):
         raise ValueError("`u` is not a pmd_unit instance")
 
+    # Delegate to power_unit(0.5): it halves the dimension exponents with true
+    # division (e.g. m -> m^0.5) and bypasses make_dimension, so fractional
+    # dimensions survive. (Building via pmd_unit(...) instead would route
+    # through make_dimension, which int()-truncates 0.5 back to 0.)
+    result = power_unit(u, 0.5)
```
electronsandstuff, line 989: What is "true division?"

Also, it isn't a good idea to bypass the constructor checks like this. If you accept floats, then you accept floats. It is also probably not a great idea to have `dimension` alternate between an array of floats and an array of ints depending on how the object was created. I would pick one.

tests/test_unit.py
```diff
@@ -17,6 +22,202 @@ def test_deprecated_unit():
         assert deprecated_unit("T") == pmd_unit("T")
 
 
+def _round_trips(u: pmd_unit) -> bool:
```
electronsandstuff, line 40: A style thing: I would do `assert_round_trip` and put the assert statement in here, then remove the try, except so the error shows up in pytest's output.

electronsandstuff (changes requested):
Looks pretty good. Dropped some comments.

beamphysics/units.py
ChristopherMayes: True division is `/` in Python, as opposed to floor division `//`: https://leapcell.io/blog/understanding-division-operators-in-python

Previously I mistakenly used floor division.

beamphysics/units.py
ChristopherMayes: refactored to use the same logic in _best_compound_match.

beamphysics/units.py
ChristopherMayes: Yes, but this is only for .simplify()> It's likely not worth optimizing or caching.

tests/test_unit.py
ChristopherMayes: added

beamphysics/units.py
ChristopherMayes: Fixed, now 7-tuple of floats. This was prescribed by the standard anyway: https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/STANDARD.md#required-for-each-record

beamphysics/units.py
ChristopherMayes: done

tests/test_labels.py
electronsandstuff, line 1: I'd refactor most into a `pytest.mark.parameterize` to make adding things like `assert mathlabel(a, units=b, tex=c) == d` easy.

beamphysics/units.py
```diff
@@ -170,6 +197,23 @@ def from_symbol(cls, unitSymbol: str) -> pmd_unit:
         ValueError
             If the symbol is not found in known_unit dict or cannot be parsed.
         """
+        # Strip outer whitespace so users can write "  eV/c" or "kg*m / s".
+        # Per-token whitespace inside the operator-split branch below is also
```
electronsandstuff: Can this comment be de-jargoned?

I don't think either of us are unicode experts and I am guessing the text talking about homoglyphs and codepoints was generated by LLM. I see what it's saying, but I am worried about LLMs being great at generating text with the veneer of competence that fails if I actually knew more about unicode parsing. Like is `folded` a technical term or just something that sounds good from the training set? I don't think that's how I would describe the conversion in plain language and I had to think for a second going through all of this.

In general, there's a lot of sections in this PR with comments / documentation that isn't wrong, but could be said in half as many words if a human touch was applied.

electronsandstuff: Also is the hardcoded translation the right way to deal with strings that map to the same unit? Could be added to the `known_units` dict as a second entry when two strings map to the same unit?

beamphysics/units.py
```diff
@@ -85,64 +98,78 @@ class pmd_unit:
...
-                self._unitSymbol = u._unitSymbol
-                self._unitSI = u._unitSI
-                self._unitDimension = u._unitDimension
+            u = self.from_symbol(unitSymbol)
+            self._unitSymbol = u._unitSymbol
+            self._unitSI = u._unitSI
+            self._unitDimension = u._unitDimension
 
     @classmethod
     def from_symbol(cls, unitSymbol: str) -> pmd_unit:
```
electronsandstuff, line 180: There is a lot of shared code between this and `_parse_single_unit`. What is the difference? Is it needed? Should we just have `from_symbol` and recurse on the broken up parts when parsing?

electronsandstuff (commented):
OK, dropped some more comments.

electronsandstuff (approved):

beamphysics/units.py
ChristopherMayes: done

beamphysics/units.py
ChristopherMayes: this all got refactored

PR #143, opened 2026-06-12
Small fixes from the units-overhaul review

ChristopherMayes:
Follow-up to #142. While reviewing every consumer of `beamphysics.units` for that PR — and in a subsequent bug sweep over the interfaces, fields, wakefields, and core modules — a collection of small bugs unrelated to the units machinery turned up. The mechanical, low-risk fixes are collected here; physics-level findings are deferred to their own PRs (see the last section).

## 1. Genesis2 `.dpa` momentum reconstruction (bug)

`beamphysics/interfaces/genesis.py::genesis2_dpa_to_data`:

```diff
- pz = np.sqrt((gamma**2 - 1) * mc2**2 - px**2 - py * 2)
+ pz = np.sqrt((gamma**2 - 1) * mc2**2 - px**2 - py**2)
```

The relation is p² = px² + py² + pz²; the typo corrupted every `pz` read from a `.dpa` file with nonzero `py`.

Also in `genesis4_par_to_data`: a `units` dict was populated from the file attributes and then never used — removed.

## 2. Twiss ignored particle weights (bug)

`beamphysics/statistics.py::particle_twiss_dispersion` computed the weighted covariance and discarded it:

```diff
- np.cov([x, xp, delta], aweights=P.weight)
- twiss = twiss_dispersion_calc(np.cov([x, xp, delta]))
+ sigma = np.cov([x, xp, delta], aweights=P.weight)
+ twiss = twiss_dispersion_calc(sigma)
```

So `ParticleGroup.twiss()` ignored weights, unlike every other statistic (`avg`, `std`, `cov` all weight). A regression test verifies that a group with half its particles down-weighted to ~zero matches the equivalent subset.

**Compatibility note:** twiss results change for unequally-weighted distributions. Equal-weight distributions are unaffected.

## 3. Statistics standard: wrong covariance units (bug)

`beamphysics/standards/statistics.py::_multiply_units` wrote `f"{unit}^2"` for equal units, so `cov_px__px` documented units of `eV/c^2` — which parses as eV/(c²), a *mass* — instead of (eV/c)². It now delegates to `pmd_unit` multiplication, matching what `pg_units` computes at runtime (`eV/c*eV/c`):

```python
def _multiply_units(unit1: str, unit2: str) -> str:
    return (pmd_unit(unit1) * pmd_unit(unit2)).unitSymbol
```

Related cleanups in the same module:

- `ARRAY_KEYS` gains the missing `y_bar py_bar` (the x-plane twins were already listed), so `sigma_y_bar`, `cov_y_bar__py_bar`, etc. now resolve. The computed-statistics standard grows from 1147 to 1287 entries; the count test now derives the expectation from `ARRAY_KEYS`/`OPERATORS` instead of hardcoding it.
- The generated markdown renders `unitSI`/`unitDimension` via `:g` (no `.0` noise from the float dimension tuples introduced in #142).
- `validate_against_particlegroup` caught `except (AttributeError, KeyError, Exception)` — i.e. everything — when probing each label, turning any failure (including bugs in the lookup itself) into a "not accessible" warning. Narrowed to `except (AttributeError, KeyError, ValueError)`: exactly the exceptions `ParticleGroup.__getitem__` raises to signal an unsupported key. Any other error now propagates loudly instead of being downgraded to a warning string.

## 4. Twiss/slice statistics correctness (bugs)

- **`ParticleGroup.twiss_match` silently dropped its `p0c` argument** (`particles.py`): it was accepted but never forwarded to `matched_particles`, so a user-supplied reference momentum was ignored. Now forwarded; regression test compares against calling `matched_particles` directly.
- **`particle_twiss_dispersion` mixed normalizations** (`statistics.py`): `xp` used the supplied `p0c` while `delta` and `norm_emit` used `mean_p`, so an explicit `p0c != mean_p` produced inconsistently-normalized dispersion and normalized emittance. All three now use `p0c` (identical results at the default `p0c=None`).
- **`slice_statistics` returned uninitialized memory** (`statistics.py`): every requested key was allocated with `np.empty`, but twiss-style keys (`twiss_x`, `twiss_xy`) are expanded into `twiss_beta_x` etc. and the raw key was never filled — the caller got garbage under `'twiss_x'`. Arrays are now allocated only for keys that get filled; regression test included.
- **`ParticleGroup.__eq__` mutated its operands and ignored species** (`particles.py`): the key list included `"id"`, and accessing `self["id"]` *assigns* fresh ids to any group lacking them — so an equality check wrote new data into both operands (and two id-less groups always compared id-equal trivially). It also never compared `species`, so an electron and positron bunch with identical arrays compared equal. Now: species is compared, ids are compared only when already present, and the check is side-effect-free. Regression test included.

## 5. Interface writer/reader fixes (bugs)

- **`write_impact` mutated the caller's beam** (`impact.py`): the cathode-start small-`pz` shift operated on the ParticleGroup's internal array (`pz[small_pz] += pz_small`), permanently modifying the input. Now operates on a copy; regression test included.
- **`write_lucretia` silently wrote nothing when `stop_ix` was given** (`lucretia.py`): a misindented `return` sat at the `if` level instead of inside it, so any caller passing a real `stop_ix` array hit the `else` and returned before `savemat` — no file, no error. The length check is now a `ValueError` and a valid `stop_ix` writes the file. Regression test included.
- **`create_fourier_data` had a dead r=0 guard** (`impact.py`): `len(np.where(...))` is the array's ndim (always 1), so the check could never fire and a missing r=0 surfaced as a bare `IndexError`. Now checks the index array itself.
- **CST reader error handling** (`cst.py`): an unknown unit string (e.g. `[cm]`) returned `None` and propagated into `array * None` far from the cause; a file with neither 6 nor 9 columns hit `NameError`. Both now raise `ValueError` at the source. Regression test included.
- **`Wavefront.from_genesis4` crashed on `h5py.Group` input** (`wavefront/wavefront.py`): the branch referenced `h5`, a name only bound in the str/Path branch (`NameError`). Now uses `file`.
- **`ansys.py` magnetic-component loop** iterated `components1` (the E-field dict) to index `components2` — works only while both parsers return identical keys; now iterates the dict it reads.
- **`gpt.py`**: the pure-electric 1D fieldmap branch computed `dat["Er"]` that was never written (dead leftover) — removed; the post-write hints said `asci2df` where the GPT tool is `asci2gdf` (4 sites).
- **`astra.py`/`superfish.py`**: `assert len(iangle) == 1, print(iangle)` — the assert message was `print`'s return value (`None`). Now a real f-string message.

## 6. FieldMesh / plot fixes (bugs)

- **Duplicate-component guard checked the wrong dict** (`fields/fieldmesh.py::load_field_data_dict`): `assert k not in data` tested the outer `{'attrs','components'}` dict, so a dict containing both `electricField/z` and its alias `Ez` silently let one overwrite the other. Review caught that the first fix only guarded the alias branch (alias-first ordering still slipped through); the loop now translates aliases to canonical names first and applies a single duplicate check, so the collision is caught regardless of dict insertion order. Regression test parametrized over both orderings.
- **Rectangular `FieldMesh.plot(..., return_figure=True)` returned `None`**: the branch didn't return, and `plot_fieldmesh_rectangular_2d` never implemented its documented `return_figure`. Both fixed (matching the cylindrical path). Review of that fix surfaced a related pre-existing break: both rectangular plotters bound `fig` only when creating their own axes, so a caller-provided `axes` hit `UnboundLocalError` (at `fig.colorbar` in the 2D plotter, at `return fig` in the 1D one). `fig` is now taken from `axes.get_figure()` in that path; regression test covers provided-axes + `return_figure`.
- **`fields/analysis.py`**: `solenoid_analysis` called `np.trapezoid` directly, bypassing the module's own numpy<2 `trapezoid`/`trapz` shim (breaks on numpy 1.x); the `axs[2,1]` diff block set zorder/patch on `axs[1,1]`/`ax202` (copy-paste from the wrong panel); the cartesian curl-E panels plotted `+Re[iωB]` but labeled it `−Re[iωB]` — label-only fix, no plotted quantity changed. The `+iωB` comparison is correct under the package's exp(−iωt) phasor convention (∇×E = −∂B/∂t → ∇×Ê = +iωB̂), the panel's own residual computes `Re[∇×E] − Re[iωB]` (the curves are meant to coincide), and the cylindrical counterpart in the same file already labels the identical expression `Re[iωB_θ]` without the minus.
- **`plot.py::slice_plot`**: `ymin = max(of per-key minima)` — for multi-key plots the lower bound should be the smallest minimum (`min`).

## 7. Validation `assert`s: messages and exceptions

`beamphysics/fields/fieldmesh.py` validated HDF5 file contents (component `unitDimension`, grid shape) with `assert`, which is stripped under `python -O` and gives poor messages. Both checks now raise `ValueError` with the expected/found values.

A follow-up sweep covered every remaining message-less `assert` in the package (19 sites). Checks that validate caller or file input — geometry requirements in the plot/expansion/conversion/interface writers, species restrictions (litrack, genesis), Genesis `.dpa`/`.par` shape checks, OPAL `ptype`, CST E/B grid consistency, on-axis-field realness (expansion, impact), and `_scalar_maybe_from_array` — now raise `ValueError` with informative messages. The two genuine internal invariants (`resample_particles` branch logic, the `testing` helper's species check) keep their `assert` but gained messages. The earlier `assert len(iangle) == 1, print(iangle)` cases (astra, superfish) — whose "message" was `print`'s return value, `None` — were fixed with real f-strings.

## 8. Docstring and comment corrections

- `particles.py`: `average_current` formula (said `*c_light*beta_z`, code divides); `norm_emit_y` said "x plane"; `n_dead` said "alive particles"; `drift_to_t` said "If no z is given"; `rotate()` documented `yc` and `zc` both as "x coordinate".
- `statistics.py`: `matched_particles` said "x, py if plane == 'y'".
- `interfaces/elegant.py`: column lists read "x, xp, y, xp" (×2).
- `interfaces/impact.py`: "GBy = gamma*beta_x" → GBx.
- `wakefields/base.py`: `convolve_density` return documented as [V]; it is [V/m].
- `wakefields/pseudomode.py`: accumulator comment said "trailing particles" where it holds the particles ahead.
- `wakefields/resistive_wall/impedance.py`: plot default documented as `100 * s0`; code uses `20 * s0`.

## 9. Test refactors

- `tests/test_labels.py` — refactored into a single `pytest.mark.parametrize` table (`mathlabel(*keys, units=..., tex=...) == expected`), per review feedback on #142; adding cases is now one line. The units-only (`no keys`) cases moved here from `tests/test_plot_labels.py`.
- `tests/test_interfaces.py` — new; regression tests for §5 (impact mutation, lucretia stop_ix, CST units).
- `tests/test_particlegroup.py` — weighted-twiss (§2), `twiss_match` p0c (§4), slice twiss keys (§4), `__eq__` species/id (§4).
- `tests/test_fieldmesh.py` — duplicate-component alias rejection (§6).
- `tests/test_statistics_standard.py` — derived count test (§3).

## Found but deferred to other PRs

Physics-level findings from the same sweep, kept out of this PR because they change numerical output and deserve dedicated tests/review:

- `fields/expansion.py`: `f0` vs `f1` term in the radial RF expansion (corrupts Er/Br for `method="fft"`; verified against the spline method); `zmirror="auto"` always mirrors (truthy string); TE-mode `Etheta` scaled by ω²/c² instead of ω.
- `fields/corrector_modeling.py` (saddle-coil cluster): `p21 = p22 + offset` typo; second arc call missing `offset`/`rotation_matrix`; `make_saddle_dipole_corrector_fieldmesh` drops `offset`/`rotation_matrix`; `rotate_around_e3` bottom-right element 0 (should be 1); hardcoded `plot_wire=True`; `if offset:` crashes on ndarray; pitch/yaw mapping contradicts docstrings.
- `interfaces/simion.py`: `identify_species` rounds `mec2` instead of `mass`; writer/reader AZ sign mismatch (round trip negates px); CWF column writes total charge instead of per-particle weight.
- `interfaces/astra.py`: probe-particle times written in seconds amid ns columns; probe `z` retains stale `-z_ref`.
- `interfaces/genesis.py`: `.dpa` z-spreading formula disagrees with its comment for `zsep != 1`.
- Remaining *messaged* `assert`s that validate file content across the interfaces (the message-less ones were converted in §7); `superfish.py` T7 reader ignores a nonzero rmin; `bunching` z-approximation sign convention differs from `apply_wakefield`.

## Files changed

- `beamphysics/interfaces/genesis.py` — §1
- `beamphysics/statistics.py` — §2, §4, §8
- `beamphysics/standards/statistics.py` — §3
- `beamphysics/particles.py` — §4, §8
- `beamphysics/interfaces/impact.py`, `lucretia.py`, `cst.py`, `ansys.py`, `gpt.py`, `astra.py`, `superfish.py`, `elegant.py`, `litrack.py`, `opal.py` — §5, §7, §8
- `beamphysics/wavefront/wavefront.py` — §5
- `beamphysics/fields/fieldmesh.py` — §6, §7
- `beamphysics/fields/analysis.py` — §6
- `beamphysics/fields/expansion.py`, `beamphysics/fields/conversion.py`, `beamphysics/testing.py` — §7
- `beamphysics/plot.py` — §6, §7
- `beamphysics/wakefields/base.py`, `pseudomode.py`, `resistive_wall/impedance.py` — §8
- `tests/test_interfaces.py` (new), `tests/test_labels.py`, `tests/test_plot_labels.py`, `tests/test_particlegroup.py`, `tests/test_fieldmesh.py`, `tests/test_statistics_standard.py` — §9

## Test results

```
1817 passed, 1 xpassed, 3 warnings in ~12s
```

## Acknowledgment

These issues were found and fixed in collaboration with **Claude** (Anthropic, via Claude Code) during and after the package-wide review for #142.

### DIFF
--- a/beamphysics/fields/fieldmesh.py
+++ b/beamphysics/fields/fieldmesh.py
@@ -1170,14 +1170,14 @@ def load_field_data_dict(data_dict, verbose=True):
     # Go through components. Allow aliases
     comp = data["components"] = {}
     for k, v in data_dict["components"].items():
-        if k in component_alias:
-            comp[k] = v
-        elif k in component_from_alias:
+        if k in component_from_alias:
             k = component_from_alias[k]
-            if k in comp:
-                raise ValueError(f"Duplicate component: {k}")
-            comp[k] = v
-        else:
+        elif k not in component_alias:
             raise ValueError(f"Unallowed component: {k}")
+        # Canonical and alias spellings of the same component collide here,
+        # regardless of which order the dict provides them in.
+        if k in comp:
+            raise ValueError(f"Duplicate component: {k}")
+        comp[k] = v
 
     return data
--- a/beamphysics/fields/fieldmesh.py
+++ b/beamphysics/fields/fieldmesh.py
@@ -557,7 +557,7 @@ class FieldMesh:
                 **kwargs,
             )
         elif self.geometry == "rectangular":
-            plot_fieldmesh_rectangular_2d(
+            return plot_fieldmesh_rectangular_2d(
                 self,
                 component=component,
                 time=time,
@@ -1129,14 +1129,17 @@ def load_field_data_h5(h5, verbose=True):
 
             # Check dimensions
             dim = h5[name].attrs["unitDimension"]
-            assert np.all(
-                dim == required_dim
-            ), f"{name} with dimension {required_dim} expected for {name}, found: {dim}"
+            if not np.all(dim == required_dim):
+                raise ValueError(
+                    f"{name}: expected unitDimension {tuple(required_dim)}, "
+                    f"found {tuple(dim)}"
+                )
 
             # Check shape
             s1 = tuple(attrs["gridSize"])
             s2 = cdat.shape
-            assert s1 == s2, f"Expected shape: {s1} != found shape: {s2}"
+            if s1 != s2:
+                raise ValueError(f"{name}: expected shape {s1}, found {s2}")
 
             # Static fields should be real
             if attrs["harmonic"] == 0:
@@ -1171,7 +1174,8 @@ def load_field_data_dict(data_dict, verbose=True):
             comp[k] = v
         elif k in component_from_alias:
             k = component_from_alias[k]
-            assert k not in data
+            if k in comp:
+                raise ValueError(f"Duplicate component: {k}")
             comp[k] = v
         else:
             raise ValueError(f"Unallowed component: {k}")
--- a/beamphysics/interfaces/astra.py
+++ b/beamphysics/interfaces/astra.py
@@ -271,7 +271,7 @@ def astra_1d_fieldmap_data(fm):
         iangle = np.unique(np.mod(np.angle(fz), np.pi))
 
         # Make sure there is only one
-        assert len(iangle) == 1, print(iangle)
+        assert len(iangle) == 1, f"Expected exactly one angle, found: {iangle}"
         iangle = iangle[0]
 
         if iangle != 0:
--- a/beamphysics/interfaces/astra.py
+++ b/beamphysics/interfaces/astra.py
@@ -466,7 +466,10 @@ def write_astra_3d_fieldmaps(fieldmesh_object, common_filePath):
         base[0:2] == "3D" or base[3:5] == "3D"
     ), "The base filename must begin with 3D or have 3D starting at the fourth character, according to Astra"
 
-    assert fieldmesh_object.geometry == "rectangular"
+    if fieldmesh_object.geometry != "rectangular":
+        raise ValueError(
+            f"Requires rectangular geometry, got: {fieldmesh_object.geometry}"
+        )
 
     nx, ny, nz = fieldmesh_object.shape
 
--- a/beamphysics/interfaces/superfish.py
+++ b/beamphysics/interfaces/superfish.py
@@ -79,7 +79,7 @@ def fish_complex_to_real_fields(fm, verbose=False):
     iangle = np.unique(np.mod(np.angle(Ez), np.pi))
 
     # Make sure there is only one
-    assert len(iangle) == 1, print(iangle)
+    assert len(iangle) == 1, f"Expected exactly one angle, found: {iangle}"
     iangle = iangle[0]
 
     if iangle != 0:
--- a/beamphysics/plot.py
+++ b/beamphysics/plot.py
@@ -156,7 +156,7 @@ def slice_plot(
         raise ValueError(f"Incompatible units: {[u.unitSymbol for u in ulist]}")
     uy = u0.unitSymbol
 
-    ymin = max([slice_dat[k].min() for k in keys])
+    ymin = min([slice_dat[k].min() for k in keys])
     ymax = max([slice_dat[k].max() for k in keys])
 
     _, f2, uy, ymin, ymax = plottable_array_and_units(
@@ -1024,6 +1024,9 @@ def plot_fieldmesh_rectangular_2d(
         orientation="vertical",
         label=llabel,
     )
+
+    if return_figure:
+        return fig
 
 
 def plot_1d_density(
--- a/beamphysics/plot.py
+++ b/beamphysics/plot.py
@@ -793,6 +793,8 @@ def plot_fieldmesh_rectangular_1d(
 
     if not axes:
         fig, axes = plt.subplots(**kwargs)
+    else:
+        fig = axes.get_figure()
 
     # Use recursion to plot multiple field components
     if isinstance(field_component, list):
@@ -906,6 +908,8 @@ def plot_fieldmesh_rectangular_2d(
 
     if not axes:
         fig, axes = plt.subplots(**kwargs)
+    else:
+        fig = axes.get_figure()
 
     if not cmap:
         cmap = CMAP1
--- a/beamphysics/plot.py
+++ b/beamphysics/plot.py
@@ -604,7 +604,8 @@ def plot_fieldmesh_cylindrical_2d(
 
     """
 
-    assert fm.geometry == "cylindrical"
+    if fm.geometry != "cylindrical":
+        raise ValueError(f"Requires cylindrical geometry, got: {fm.geometry}")
 
     if mirror not in (None, "r"):
         raise ValueError("mirror must be None or 'r'")
@@ -890,7 +891,8 @@ def plot_fieldmesh_rectangular_2d(
 
     """
 
-    assert fm.geometry == "rectangular"
+    if fm.geometry != "rectangular":
+        raise ValueError(f"Requires rectangular geometry, got: {fm.geometry}")
 
     valid_coordinates = set(fm.axis_labels)
     coordinate_value = None
--- a/beamphysics/standards/statistics.py
+++ b/beamphysics/standards/statistics.py
@@ -63,7 +63,7 @@ p energy kinetic_energy xp yp higher_order_energy
 r theta pr ptheta
 Lz
 gamma beta beta_x beta_y beta_z
-x_bar px_bar Jx Jy
+x_bar px_bar y_bar py_bar Jx Jy
 """.split()
 
 # Operators and their properties for computed statistics
@@ -260,7 +260,7 @@ def validate_against_particlegroup(standard: Dict[str, Any]) -> List[str]:
         # Try to access the attribute
         try:
             value = P[label]
-        except (AttributeError, KeyError, Exception) as e:
+        except Exception as e:
             warnings.append(f"Label '{label}' not accessible in ParticleGroup: {e}")
             continue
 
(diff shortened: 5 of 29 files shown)


beamphysics/plot.py
```diff
@@ -1025,6 +1025,9 @@ def plot_fieldmesh_rectangular_2d(
         label=llabel,
     )
 
+    if return_figure:
+        return fig
```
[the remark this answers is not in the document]
ChristopherMayes: fixed in a different way

beamphysics/fields/fieldmesh.py
```diff
@@ -1171,7 +1174,8 @@ def load_field_data_dict(data_dict, verbose=True):
             comp[k] = v
         elif k in component_from_alias:
```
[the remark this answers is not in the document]
ChristopherMayes: fixed

tests/test_fieldmesh.py
```diff
@@ -66,3 +66,28 @@ def test_fieldmesh_limits(fm_dipole_corrector):
...
+        "gridSize": (2, 2, 2),
+        "harmonic": 0,
+        "fundamentalFrequency": 0,
+    }
+    data = {
+        "attrs": attrs,
+        "components": {"electricField/z": arr, "Ez": 2 * arr},
+    }
+    with pytest.raises(ValueError, match="Duplicate component"):
+        load_field_data_dict(data, verbose=False)
```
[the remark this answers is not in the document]
ChristopherMayes: fixed

beamphysics/interfaces/astra.py
```diff
@@ -271,7 +271,7 @@ def astra_1d_fieldmap_data(fm):
         iangle = np.unique(np.mod(np.angle(fz), np.pi))
 
         # Make sure there is only one
-        assert len(iangle) == 1, print(iangle)
+        assert len(iangle) == 1, f"Expected exactly one angle, found: {iangle}"
```
electronsandstuff, line 274: This should be exception, not assert

electronsandstuff: Maybe just search for all asserts and replace

beamphysics/interfaces/superfish.py
```diff
@@ -79,7 +79,7 @@ def fish_complex_to_real_fields(fm, verbose=False):
     iangle = np.unique(np.mod(np.angle(Ez), np.pi))
 
     # Make sure there is only one
-    assert len(iangle) == 1, print(iangle)
+    assert len(iangle) == 1, f"Expected exactly one angle, found: {iangle}"
```
electronsandstuff, line 82: Exception

beamphysics/standards/statistics.py
```diff
@@ -260,7 +260,7 @@ def validate_against_particlegroup(standard: Dict[str, Any]) -> List[str]:
         # Try to access the attribute
         try:
             value = P[label]
-        except (AttributeError, KeyError, Exception) as e:
+        except Exception as e:
```
electronsandstuff: Avoid general try:... except:... blocks, the old version only catching `KeyError` and `AttributeError` was better to not hide issues.

electronsandstuff (commented):
Looks good. Dropped some comments.

beamphysics/standards/statistics.py
ChristopherMayes: fixed

beamphysics/interfaces/astra.py
ChristopherMayes: Many assertions changed to exceptions

beamphysics/interfaces/superfish.py
ChristopherMayes: Many assertions changed to exceptions

PR #144, opened 2026-06-16
Use default_id in equality check

ChristopherMayes:
Previously two ParticleGroups `__eq__` comparison would return False if one had an id array = 1-N, which is the default, and the other didn't have this assigned. This checks against the default for differences. 

Claude notes:

## Summary

`ParticleGroup.__eq__` previously returned `False` whenever one operand had an
explicit `id` array and the other did not — even when the operand without ids
would, on access, be assigned the default ids `1..n` that exactly matched the
other operand. This made two otherwise-identical particle groups compare unequal
purely because of how ids were stored.

## Changes

- **`beamphysics/particles.py`**
  - Added a module-level `default_id(n)` helper that returns the default
    particle ids (`np.arange(1, n + 1)`).
  - Updated `assign_id` to use `default_id` for consistency.
  - Reworked `__eq__`:
    - Compares all non-id arrays (`x`, `px`, `y`, `py`, `z`, `pz`, `t`,
      `status`, `weight`) first.
    - For ids, if only one operand has an explicit `id` array, the existing id
      is compared against the default `1..n` ids the other operand would receive
      from `assign_id`.
    - Avoids triggering `assign_id` as a side effect (accessing `self["id"]`
      would otherwise assign fresh ids to both operands).

## Behavior

| Left ids        | Right ids       | Result                                    |
| --------------- | --------------- | ----------------------------------------- |
| none            | none            | equal (ids not compared)                  |
| `1..n` default  | none            | **equal** (was previously unequal)        |
| explicit custom | none            | equal only if custom ids == `1..n`        |
| explicit        | explicit        | equal only if id arrays match             |

## Notes

- No side effects: comparison no longer mutates either operand by assigning ids.
- Docstrings follow numpydoc style.

### DIFF
--- a/beamphysics/particles.py
+++ b/beamphysics/particles.py
@@ -1849,8 +1849,7 @@ class ParticleGroup:
             if self_has_id or other_has_id:
                 self_id = self._data["id"] if self_has_id else default_id(len(self))
                 other_id = other._data["id"] if other_has_id else default_id(len(other))
-                if not np.array_equal(self_id, other_id):
-                    return False
+                return np.array_equal(self_id, other_id)
             return True
 
         return NotImplemented
--- a/beamphysics/particles.py
+++ b/beamphysics/particles.py
@@ -387,7 +387,7 @@ class ParticleGroup:
         """
         if "id" not in self._settable_array_keys:
             self._settable_array_keys.append("id")
-        self.id = np.arange(1, self["n_particle"] + 1)
+        self.id = default_id(self["n_particle"])
 
     @property
     def n_particle(self):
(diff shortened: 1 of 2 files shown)


beamphysics/particles.py
```diff
@@ -1836,16 +1836,21 @@ def __eq__(self, other):
...
+            # Compare ids without triggering assign_id as a side effect. If only
+            # one operand has explicit ids, compare them against the default ids
+            # (1..n) that assign_id would otherwise produce for the other.
+            self_has_id = "id" in self._data
+            other_has_id = "id" in other._data
+            if self_has_id or other_has_id:
+                self_id = self._data["id"] if self_has_id else default_id(len(self))
+                other_id = other._data["id"] if other_has_id else default_id(len(other))
+                if not np.array_equal(self_id, other_id):
+                    return False
```
ken-lauer:
```suggestion
                return np.array_equal(self_id, other_id)
```

ken-lauer:
A couple tests to validate both scenarios would probably be worthwhile

ken-lauer (approved):

PR #147, opened 2026-07-01
FIX: ParticleGroup id/status must be integer arrays

ken-lauer:
* Coerce `id`/`status` into `dtype=int`
* Other settable arrays get `dtype=float`

`asarray()` gives you back the original if it's compatible with the requested `dtype`, so I think filtering all of the data through it is reasonable.


ChristopherMayes:
Casting floats will truncate to zero (0.99999 -> 0), so perhaps it needs calls to `round`

ChristopherMayes (approved):

PR #152, opened 2026-08-14
Add `t_offset` Argument to `ParticleGroup.write` 

electronsandstuff:
This PR adds the new argument `t_offset`  to `ParticleGroup.write`. This gets written as the OpenPMD `timeOffset` attribute to avoid numerical roundoff error when writing `ParticleGroup` objects with reference time included for some codes (ie Bmad which reads them separately and stores internally as a quad precision floating point number.

Full changes:
 - New argument `t_offset` in `ParticleGroup.write` supporting scalar and array types. Does not write `timeOffset` when not specified.
 - Update `write_pmd_bunch` docstring to numpy
 - Add cell that demonstrates use of `t_offset` to `docs/examples/write_examples.ipynb`

## Screenshot
image

### DIFF
--- a/beamphysics/particles.py
+++ b/beamphysics/particles.py
@@ -1415,7 +1415,7 @@ class ParticleGroup:
         return write_opal(self, filePath, verbose=verbose, dist_type=dist_type)
 
     # openPMD
-    def write(self, h5, name=None) -> None:
+    def write(self, h5, name=None, t_offset=0.0) -> None:
         """
         Write particle data to an HDF5 file or group in openPMD format.
 
@@ -1433,6 +1433,9 @@ class ParticleGroup:
         name : str, optional
             Name for the subgroup/bunch written inside the "particles" group (or provided group).
             If None, `write_pmd_bunch` will write directly to the "particles" group.
+        t_offset : float or numpy.ndarray, optional
+            Time offset, scalar or per-particle, written as the openPMD "timeOffset"
+            record. Omitted when zero. Default is 0.0.
 
         Returns
         -------
@@ -1465,7 +1468,7 @@ class ParticleGroup:
         else:
             g = h5
 
-        write_pmd_bunch(g, self, name=name)
+        write_pmd_bunch(g, self, name=name, t_offset=t_offset)
 
     # Plotting
     # --------
--- a/beamphysics/particles.py
+++ b/beamphysics/particles.py
@@ -2268,7 +2268,7 @@ def _scalar_maybe_from_array(value):
     return value[0]
 
 
-def load_bunch_data(h5):
+def load_bunch_data(h5, include_offset=True):
     """
     Load particles into structured numpy array.
     """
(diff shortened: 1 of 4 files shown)


beamphysics/writers.py
```diff
@@ -38,19 +38,27 @@ def pmd_field_init(h5, externalFieldPath="/ExternalFieldPath/%T/"):
...
+    h5 : h5py.File or h5py.Group
+        Handle to write into.
+    data : dict or ParticleGroup
+        Requires keys 'x', 'px', 'y', 'py', 'z', 'pz', 't', 'status', 'weight'
+        (arrays), 'species' (str), 'n_particle' (int), 'charge' (float).
+        Optional key: 'id' (array).
+    name : str, optional
+        Subgroup to create for the bunch. If None, writes directly into `h5`.
+    t_offset : float or numpy.ndarray, optional
+        Time offset, scalar or per-particle, written as the 'timeOffset' record.
```
ken-lauer, line 56: Per-particle offset seems a strange thing to me (as clearly a non-physicist), but apparently it's a reasonable thing?
[The standard](https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_BeamPhysics.md#per-particle-records-of-the-particle-group) does indicate this, after all:
> The reference time may depend upon the longitudinal position of a particle and so may be different for different particles.

beamphysics/particles.py
ken-lauer, line 1: The time offset itself seems to be special enough to deserve its own spot in `ParticleGroup`. Well, at least from my usual standpoint of "if your data isn't round-trippable (de/serializable) to its original representation it's a problem".

Trying to think this through a bit:

```python
# assume timeOffset is set in particles.h5
P = ParticleGroup("particles.h5")   # defaults to include_offset=True
P.t  # includes timeOffset from the original file
# P.time_offset ❌ not retrievable or otherwise inspectable
P.write("out.h5", t_offset=0.0)  # not round-trippable / lossy: timeOffset goes away
```

The test suite does:

```python
        ParticleGroup(
            data=load_bunch_data(hf[particle_paths(hf)[0]], include_offset=False)
        )
```

And then allows it to be handled on the side. This `include_offset` flag opts out of _all_ offsets wholesale though, so you can't just opt out of time offset handling. A bit awkward, I think.

ken-lauer (commented):
If you and @ChristopherMayes say the round-trippability doesn't matter, then I'm fine with this proceeding as-is.

ken-lauer (approved):
Well, I don't like it but @ChristopherMayes said it's the design intent. 

In short, LGTM.

beamphysics/particles.py
electronsandstuff: Yeah, I think it's up to @ChristopherMayes if he wants to start including the offsets. I agree with you on being able to round-trip in principle. This was the smallest atomic edit I could make to avoid the numerical artifacts without doing a big architecture change of `ParticleGroup`.

If we wanted to think about a bigger change, one path forward could be to promote `x`, `y`, `z`, `t`... to properties and have the actual fields be `raw_x`, `offset_x` with `.x` returning `raw_x + offset_x`. This would keep the current behavior while supporting the offset fields (which are in the OpenPMD standard). For the sake of small PRs and keeping work flowing, it might be best to start here and think about the bigger `ParticleGroup` change in a separate issue?

@ChristopherMayes?

edit: Just read below comment, I will merge and add the offset notes as a suggestion in an issue

beamphysics/writers.py
electronsandstuff: Yes, I allowed offsets because it's in the standard. I also don't have an immediate use, but if it's allowed in the standard, we might as well expose it.

